PR #3, opened 2018-05-09
Move individual formula pages to /formula

koraktor:
This matches the paths of the old Rails app. We should not break existing links.

As a side-effect, this also saves us from breakage if a formula called `index` was introduced.

### DIFF
--- a/_data/formula/ansible@1.9.json
+++ b/_data/formula/ansible@1.9.json
@@ -48,22 +48,22 @@
   ],
   "bottle": {
     "stable": {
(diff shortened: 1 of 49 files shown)


MikeMcQuaid:
Ideally can you split the directory rename and other changes in here into separate commits? It'll make it easier to figure out what's going on for review, thanks!

koraktor:
Ok, that’s probably the best way to handle these big commits.

I removed `/formulae/index.html` by mistake. Fixed.

PR #16, opened 2018-05-23
set custom link

ldct:
Will kick in after https://github.com/Homebrew/brew.sh/pull/222 lands

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -29,6 +29,9 @@ defaults:
 
 logo: https://brew.sh/assets/img/homebrew-256x256.png
 
+github:
+  forkme_repo: Homebrew/formulae.brew.sh
+
 twitter:
   username: MacHomebrew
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
@ldct It should be possible to use https://help.github.com/articles/repository-metadata-on-github-pages/ rather than hardcoding this. Thoughts?

ldct:
I think the problem would be that the site currently hosted on brew.sh, which is running code from the repo Homebrew/brew.sh, wants to point to the repo Homebrew/brew

MikeMcQuaid:
Could use it for formulae.brew.sh (and docs.brew.sh) though, no?

ldct:
yes, but you'd need to override it for brew.sh

ldct:
also it's not clear to me which repo https://docs.brew.sh/ actually wants to point users to

MikeMcQuaid:
> also it's not clear to me which repo https://docs.brew.sh/ actually wants to point users to

I've made it link to the individual pages for easy editing.

MikeMcQuaid:
Thanks again @ldct!

Issue #13, opened 2018-05-11
Add Casks


MikeMcQuaid:
We should consider including Homebrew Casks into formulae.brew.sh. This would require generating an equivalent of `brew info --json=v1` for Casks on a Linux machine.

(Imported from https://github.com/Homebrew/formulae.brew.sh/issues/117 CC @adidalal)

MikeMcQuaid:
> brew cask info {{cask_name}} does give you all the necessary details about a cask, and there's also brew cask _stanza {{stanza_name}} {{cask_name}} for specific bits.

Do either of them output JSON? If not, coming up with a format and building that on the Cask side would be a prerequisite.

adidalal:
From my limited knowledge, I don't believe so.
/cc @reitermarkus @vitorgalvao for visibility

reitermarkus:
`brew cask _stanza` has a `--yaml` option.

MikeMcQuaid:
The Homebrew/brew side has been implemented in https://github.com/Homebrew/brew/pull/4410

PatMyron:
From these:
https://github.com/Homebrew/brew.sh/pull/228
https://github.com/Homebrew/brew/pull/4620

It looks like @brianmorton is currently working on this, right?

Do you want any additional help? I'd love to see Cask analytics on the website

MikeMcQuaid:
@PatMyron We've not seen any PRs and I've not had any additional communication there so I'd suggest you start your own work on this.

brianmorton:
@PatMyron I am indeed mired in some stuff and help is always welcome :)

PatMyron:
@luzpaz actually just switched jobs and moved myself, so I probably won't be making much progress on this anytime soon

cmbernard333:
I would love to see this implemented. How can I get started to help contribute to this?

brianmorton:
I am actually picking this back up this week (today).

MikeMcQuaid:
I'd encourage others (e.g. @cmbernard333) to start their own implementations of this; I don't think it's worth blocking on a single individual.

brianmorton:
@MikeMcQuaid I couldn't agree more!

I have actually just been trying to read back through all the previous threads and PRs to see what needs to be done. I think it might be helpful to put that information down here as it's a bit unclear.

The goal (as I currently understand it) is to update https://formulae.brew.sh/analytics/ to include cask analytics. That will necessitate updating various files in this repository.

Have I missed anything?

[...]

PR #92, opened 2019-01-24
add cask install analytics

brianmorton:
https://github.com/Homebrew/formulae.brew.sh/issues/13

edit: also requires https://github.com/Homebrew/homebrew-formula-analytics/pull/59 to be implemented in order to generate new/updated analytics

PR #93, opened 2019-01-29
Update generate.rb to create json files for casks in new directories for API

cmbernard333:
#13 

Add json generation to `generate.rb` for casks.

This does **not** generate analytics or HTML as this will be addressed in a separate PR.

### DIFF
--- a/script/generate.rb
+++ b/script/generate.rb
@@ -1,5 +1,6 @@
 #!/usr/bin/env brew ruby
-directories = ["_data/formula", "api/formula", "formula"]
+require 'cask/cask'
+directories = ["_data/formula", "api/formula", "formula", "_data/cask", "api/cask", "cask"]
 FileUtils.rm_rf directories
 FileUtils.mkdir_p directories
 
@@ -11,4 +12,15 @@ Formula.each do |f|
   html = IO.read "_formula.html.in"
   html.gsub!("title: $TITLE", "title: #{f.name}")
   IO.write("formula/#{f.name}.html", html)
+end
+
+Cask::Cask.each do |c|
+  json_filename = "#{File.basename(c.sourcefile_path,'.rb')}.json"
+  IO.write("_data/cask/#{json_filename}", JSON.pretty_generate(c.to_h))
+  FileUtils.cp "_api_formula.json.in", "api/cask/#{json_filename}"
+
+  # TODO: generate cask html template and output(s)
+  # html = IO.read "_formula.html.in"
+  # html.gsub!("title: $TITLE", "title: #{c.name}")
+  # IO.write("cask/#{c.name}.html", html)
 end
--- a/script/generate.rb
+++ b/script/generate.rb
@@ -1,5 +1,4 @@
 #!/usr/bin/env brew ruby
-require 'cask/cask'
 directories = ["_data/formula", "api/formula", "formula", "_data/cask", "api/cask", "cask"]
 FileUtils.rm_rf directories
 FileUtils.mkdir_p directories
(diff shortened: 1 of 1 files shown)


script/generate.rb
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env brew ruby
-directories = ["_data/formula", "api/formula", "formula"]
+require 'cask/cask'
```
MikeMcQuaid: Maybe move the require directory about the `Cask::Cask.each`?

script/generate.rb
```diff
@@ -12,3 +13,14 @@
   html.gsub!("title: $TITLE", "title: #{f.name}")
   IO.write("formula/#{f.name}.html", html)
 end
+
+Cask::Cask.each do |c|
+  json_filename = "#{File.basename(c.sourcefile_path,'.rb')}.json"
+  IO.write("_data/cask/#{json_filename}", JSON.pretty_generate(c.to_h))
+  FileUtils.cp "_api_formula.json.in", "api/cask/#{json_filename}"
```
MikeMcQuaid: Does this file work perfectly? If so, can it be renamed to `_api.json.in` instead? Thanks!

script/generate.rb
```diff
@@ -12,3 +13,14 @@
...
+
+  # TODO: generate cask html template and output(s)
+  # html = IO.read "_formula.html.in"
+  # html.gsub!("title: $TITLE", "title: #{c.name}")
+  # IO.write("cask/#{c.name}.html", html)
```
MikeMcQuaid: Can remove the commented-out, non-TODO lines.

script/generate.rb
cmbernard333: Yes it does work (see below). Should this also be adjusted for formula?

```
christianbernard@Dirac-MBP formulae.brew.sh (feature/formulae.brew.sh-generate-cask-json) $ cat _data/cask/chromium.json
{
  "name": [
    "Chromium"
  ],
  "homepage": "https://www.chromium.org/Home",
  "url": "https://download-chromium.appspot.com/dl/Mac?type=snapshots",
  "appcast": null,
  "version": "latest",
  "sha256": "no_check",
  "artifacts": [
    [
      "chrome-mac/Chromium.app"
    ],
    {
      "trash": [
        "~/Library/Preferences/org.chromium.Chromium.plist",
        "~/Library/Caches/Chromium",
        "~/Library/Application Support/Chromium",
        "~/Library/Saved Application State/org.chromium.Chromium.savedState"
      ],
      "signal": [

      ]
    }
  ],
  "caveats": "",
  "depends_on": {
  },
  "conflicts_with": {
  },
  "container": null,
  "auto_updates": null
```

script/generate.rb
cmbernard333: Will do.

script/generate.rb
MikeMcQuaid: Seems good 👍

MikeMcQuaid:
Thanks @cmbernard333, nice work! Should generate this on the next homebrew-core push. Will need to add a similar thing to https://github.com/Homebrew/homebrew-core/blob/master/.travis.yml to homebrew-cask

cmbernard333:
@MikeMcQuaid Are we waiting on analytics as well (PR #92) so we can do analytics generation after creating the JSON for the cask API? Or should I just create a PR to call the script in the .travis.yml inside homebrew-cask?

MikeMcQuaid:
@cmbernard333 The analytics are done every 24h rather than on every push. Feel free to pick anything up that's in progress and cherry-pick commits. Thanks!

MikeMcQuaid:
@cmbernard333 These are not being generated: https://github.com/Homebrew/formulae.brew.sh/tree/master/_data

cmbernard333:
@MikeMcQuaid per which build? The homebrew-cask build?

MikeMcQuaid:
@cmbernard333 my expectation from this PR was that the generate script would start committing those files but this has not happened (see https://github.com/Homebrew/formulae.brew.sh/tree/master/_data). Perhaps they need to be manually committed first.

cmbernard333:
That makes sense. Should I create a PR for this repo with the generated homebrew casks based on the latest set of casks?

cmbernard333:
Here is a PR adding the initial set of json for casks in `_data/cask` - this does not include the HTML yet

MikeMcQuaid:
I'm now thinking it may be worth having a separate generate script that's run for formulae and for casks as otherwise we'll need to make sure both repos are up to date to generate either. Separate jobs mean we can just run one of them from each relevant CI (homebrew-core and homebrew-cask).

cmbernard333:
@MikeMcQuaid Roger. I'll make a separate PR and make a new script `generate-cask.rb`.

PR #99, opened 2019-02-01
Separate teh formulae and cask generation tasks in the Rakefile.

cmbernard333:
Separating out generation of JSON for formulae and cask into their own scripts and targets.


script/generate-cask.rb
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env brew ruby
+require 'cask/cask'
+directories = ["_data/cask", "api/cask", "cask"]
+FileUtils.rm_rf directories
+FileUtils.mkdir_p directories
+
+Cask::Cask.each do |c|
+  json_filename = "#{File.basename(c.sourcefile_path,'.rb')}.json"
```
reitermarkus: Why `File.basename(c.sourcefile_path,'.rb')`? Isn't this the same as `c.token`?

script/generate-cask.rb
```diff
@@ -0,0 +1,11 @@
...
+  IO.write("_data/cask/#{json_filename}", JSON.pretty_generate(c.to_h))
+  FileUtils.cp "_api_formula.json.in", "api/cask/#{json_filename}"
```
MikeMcQuaid: Rename this file now maybe?

script/generate-cask.rb
cmbernard333: You are correct. I did not see this. I am updating it in my next commit.

script/generate-cask.rb
cmbernard333: Like create a new file `_api_cask.json.in`? The current file is a template that exists in the repo.

MikeMcQuaid: Yes, either a new file or rename the other to just `_api.json.in` to reflect that it's used for formulae and casks.

cmbernard333: I'm not sure where the json is served - is the variable `formula_json` going to be accurately filled for casks? Or would we need to change it to `cask_json`?

MikeMcQuaid: I suggest testing it locally and seeing what works. New files may be needed and when they are not it's worth renaming anything that references a formula to cask instead.

cmbernard333: I created a `_layouts/cask_json.json` [gist](https://gist.github.com/cmbernard333/3a3d445458bb15128dc86ffd2d6c4744) and updated it to pull the correct information in `_data/cask/<cask>` but it seems json elements in the analytics are not named 

```
    {
      "number": 9,
      "cask": "google-chrome",
      "count": "11,375",
      "percent": "1.76"
    },
```

So it will not be able to pull in the analytics information.

MikeMcQuaid: @cmbernard333 That's done by a separate analytics job for homebrew-core: https://github.com/Homebrew/formulae.brew.sh/blob/master/Rakefile#L58-L66

It may make sense to just skip it for now.

cmbernard333: Ok. That should be it then. I need to create an associated PR for `homebrew-cask` to pull down and run `rake cask`.

cmbernard333: Here's the PR [#58112](https://github.com/Homebrew/homebrew-cask/pull/58112) for `hombrew-cask` but it can't be updated to use the new rake target until this PR is merged.

_layouts/cask_json.json
```diff
@@ -0,0 +1,95 @@
+---
+---
+{%- assign full_name = page.name | remove: ".json" -%}
+{%- assign name = full_name | remove: "@" | remove: "." | remove: '+' -%}
```
MikeMcQuaid: Are there any casks with `@` or `+` in their names?

_layouts/cask_json.json
```diff
@@ -0,0 +1,95 @@
...
+    {%- if forloop.last == false -%}
+    ,
+    {%- endif -%}
+  {%- endfor -%}
+{%- else -%}
+  {{ full_name | jsonify }}:0
+{%- endif -%}
+}},"install_on_request":{"30d":{
+
+{%- if site.data.analytics.install-on-request.homebrew-core["30d"].items[full_name].size > 0 -%}
```
MikeMcQuaid: This will never find an item as it's only for formulae.

Rakefile
```diff
@@ -11,6 +11,12 @@ task :formulae do
   sh "brew", "ruby", "script/generate.rb"
 end
 
+desc "Dump cask data"
+task :cask do
```
MikeMcQuaid: `task build: :dump` below needs to be updated to `task build: [:dump, :cask]`. `:dump` should arguably be renamed to `:formulae_and_analytics` if it's no longer dumping everything.

_layouts/cask_json.json
cmbernard333: Does this include all the analytics for `install-on-request`? If so I'll just take them all out.

_layouts/cask_json.json
cmbernard333: I don't believe so. I just pulled this from formula_json in case there were.

_layouts/cask_json.json
cmbernard333: Removed.

Rakefile
cmbernard333: Updated.

cmbernard333:
@MikeMcQuaid What are the next steps here?

_layouts/cask_json.json
MikeMcQuaid: @cmbernard333 Can you check? If there's none: these can be removed.

cmbernard333:
@MikeMcQuaid quick search `brew search --casks | egrep -i "@|\.|\+"` shows nothing. I'll take it out.

cmbernard333:
@MikeMcQuaid Here is testing of it locally https://asciinema.org/a/i12nVPfVIe2724gE2PW5YLXn9 - looks good from my stand point.

MikeMcQuaid:
Looks good, thanks again @cmbernard333

PR #111, opened 2019-03-06
Added newlines to generated cask json.

cmbernard333:


script/generate-cask.rb
```diff
@@ -7,5 +7,6 @@
 Cask::Cask.each do |c|
   json_filename = "#{c.token}.json"
   IO.write("_data/cask/#{json_filename}", JSON.pretty_generate(c.to_h))
```
MikeMcQuaid: Could this use `IO.puts` instead? If not, how about doing `#{JSON.pretty_generate(c.to_h)}\n`?

cmbernard333: Whatever makes the most sense. I was just using my limited ruby knowledge here. XD

script/generate-cask.rb
```diff
@@ -6,6 +6,6 @@
 
 Cask::Cask.each do |c|
   json_filename = "#{c.token}.json"
-  IO.write("_data/cask/#{json_filename}", JSON.pretty_generate(c.to_h))
+  IO.write("_data/cask/#{json_filename}", {JSON.pretty_generate(c.to_h)}\n)
```
reitermarkus: This isn't valid syntax, should be

```ruby
"#{JSON.pretty_generate(c.to_h)}\n"
```

MikeMcQuaid:
Thanks @cmbernard333!

PR #114, opened 2019-03-19
add support for Casks

EricFromCanada:
This adds the files necessary to add a listing of all Casks at `/cask/` and `/api/cask.json`, individual Cask pages at `/cask/{token}`, and the contents of `/cask/` via `generate-cask.rb`. (Another PR is still needed to add the files generated with `rake`.) Also fixes a few small issues with the text and template markup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 .bundle/
 _site/
 bin/
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [Homebrew Formulae](https://formulae.brew.sh) is an online package browser for [Homebrew](https://brew.sh).
 
-It displays all packages in [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core). It is deployed to GitHub Pages after every [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) commit by [Travis CI](https://travis-ci.org).
+It displays all packages in [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) and [Homebrew/homebrew-cask](https://github.com/Homebrew/homebrew-cask). A GitHub Action in [homebrew-core](https://github.com/Homebrew/homebrew-core/blob/master/.github/main.workflow) or [homebrew-cask](https://github.com/Homebrew/homebrew-cask/blob/master/.github/main.workflow) is run on each commit, which deploys the site to GitHub Pages.
 
 ## JSON API
 
@@ -19,7 +19,7 @@ Read more in the [JSON API documentation](https://formulae.brew.sh/docs/api/).
 ## Usage
 Open https://formulae.brew.sh/ in your web browser.
 
-Instead, to run Homebrew Formulae locally run:
+To instead run Homebrew Formulae locally, run:
 ```bash
 git clone https://github.com/Homebrew/formulae.brew.sh
 cd formulae.brew.sh
--- a/_includes/cask.html
+++ b/_includes/cask.html
@@ -1,0 +1,3 @@
+<td><a href="{{ site.baseurl }}/cask/{{ include.token }}">{{ include.token }}</a></td>
+<td>{{ include.cask.name.first }}</td>
+<td>{{ include.cask.version | truncate: 25 }}</td>
--- a/_layouts/analytics.html
+++ b/_layouts/analytics.html
@@ -20,7 +20,9 @@ layout: base
     {%- if page.category == "os-version" -%}
       <code>{{ item.os_version }}</code>
     {%- elsif page.category == "cask-install" -%}
-      <code>{{ item.cask }}</code>
+      <a href="{{ site.baseurl }}/cask/{{ item.cask }}">
+        <code>{{ item.cask }}</code>
+      </a>
     {%- else -%}
       {%- assign name = item.formula | split: " " | first -%}
       {%- assign data_name = name | remove: "@" | remove: "." | remove: '+' -%}
--- a/_layouts/analytics.html
+++ b/_layouts/analytics.html
@@ -1,7 +1,7 @@
 ---
 layout: base
 ---
-{% assign json = site.data.analytics[page.category][page.days] %}
+{%- assign json = site.data.analytics[page.category][page.days] -%}
 <h2>{{ page.category_pretty }} Events</h2>
 
 <h3><a href="{{ site.baseurl }}/api/analytics/{{ page.category }}/{{ page.days}}.json"><code>/api/analytics/{{ page.category }}/{{ page.days}}.json</code> (JSON API)</a></h3>
--- a/cask_index.html
+++ b/cask_index.html
@@ -1,0 +1,19 @@
+---
+layout: default
+permalink: /cask/
+---
+<p>This is a listing of all Casks available via the <a href="https://brew.sh">Homebrew</a> package manager for macOS.</p>
+
+<h2><a href="{{ site.baseurl }}/api/cask.json"><code>/api/cask.json</code> (JSON API)</a></h2>
+
+<table>
+    {%- assign sorted_casks = site.data.cask | sort -%}
+    {%- for cask in sorted_casks -%}
+
+    <tr>
+        {%- assign t = cask[0] -%}
+        {%- assign c = cask[1] -%}
+        {%- include cask.html token=t cask=c -%}
+    </tr>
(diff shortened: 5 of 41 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 [Homebrew Formulae](https://formulae.brew.sh) is an online package browser for [Homebrew](https://brew.sh).
 
-It displays all packages in [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core). It is deployed to GitHub Pages after every [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) commit by [Travis CI](https://travis-ci.org).
+It displays all packages in [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) and [Homebrew/homebrew-cask](https://github.com/Homebrew/homebrew-cask). It is deployed to GitHub Pages after every [Homebrew/homebrew-core](https://github.com/Homebrew/homebrew-core) and [Homebrew/homebrew-cask](https://github.com/Homebrew/homebrew-cask) commit by [Brew Test Bot](https://docs.brew.sh/Brew-Test-Bot).
```
MikeMcQuaid: Thanks! Could this link to https://github.com/Homebrew/homebrew-core/blob/master/.github/main.workflow and https://github.com/Homebrew/homebrew-cask/blob/master/.github/main.workflow and state that GitHub Actions makes the commits?

_includes/cask.html
```diff
@@ -0,0 +1,3 @@
+<td><a href="{{ site.baseurl }}/cask/{{ include.token }}">{{ include.token }}</a></td>
+<td>{{ include.cask.name.first }}</td>
+<td>{{ include.cask.version | truncate: 25 }}</td>
```
MikeMcQuaid: Why truncate this, out of interest?

MikeMcQuaid:
> Another PR is still needed to add the files generated with `rake`

Can you elaborate on this? Basically just "another PR is needed to add all the files that would be generated"? I believe that'll happen as soon as this PR is merged; could you maybe generate a single file and we could see the output for review?

EricFromCanada:
The `rake` task generates hundreds of new files that I didn't want crowding out the changes above, but I'll add a bunch of them as a preview.

MikeMcQuaid:
> but I'll add a bunch of them as a preview.

Thanks! Shout when done.

MikeMcQuaid:
I'm getting this trying to run generation locally:

```
brew ruby script/generate-cask.rb
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/lib/ruby/2.3.0/open-uri.rb:225:in `open_loop': redirection forbidden: http://downloads.kicad-pcb.org/osx/nightly/ -> https://kicad-downloads.s3.cern.ch/index.html (RuntimeError)
```

EricFromCanada:
Where might I find info on what formatting and contents `formula-analytics` expects to find in `.homebrew_analytics.json`?

MikeMcQuaid:
> Where might I find info on what formatting and contents `formula-analytics` expects to find in `.homebrew_analytics.json`?

@EricFromCanada You shouldn't need to set that stuff up for this. You can set `HOMEBREW_NO_ANALYTICS` when generating or just run `rake cask` to be sure.

EricFromCanada:
@MikeMcQuaid I've added some sample files for Casks.

That error is from `kicad-nightly` which is in `homebrew/cask-versions`; it's because `scripts/generate-cask.rb` generates data for every Cask in every local tap, rather than just what's in `homebrew/cask`; same for `scripts/generate.rb` with formulae. Not sure of the best way to mod the scripts to filter those.

MikeMcQuaid:
> That error is from `kicad-nightly` which is in `homebrew/cask-versions`

Gotcha, can be ignored then.

> Not sure of the best way to mod the scripts to filter those.

Could use `Tap#formula_names` or `Tap#cask_files`

> I've added some sample files for Casks.

Great, thanks, will take a look.

EricFromCanada:
> Could use Tap#formula_names or Tap#cask_files

Good idea, added.

cmbernard333:
@EricFromCanada awesome work. It will be nice to see all this reachable from the API.

script/generate-cask.rb
```diff
@@ -4,8 +4,13 @@
...
 
-Cask::Cask.each do |c|
+Tap.default_cask_tap.cask_files.each do |p|
+  c = Cask::CaskLoader::FromPathLoader.new(p).load
   json_filename = "#{c.token}.json"
   IO.write("_data/cask/#{json_filename}", "#{JSON.pretty_generate(c.to_h)}\n")
   FileUtils.cp "_api_cask.json.in", "api/cask/#{json_filename}"
+
+  html = IO.read "_cask.html.in"
+  html.gsub!("title: $TITLE", "title: \"#{c.token}\"")
```
EricFromCanada: Had to add quotes here because the `0xED` Cask page was showing `237` anywhere that `page.title` was used.

MikeMcQuaid:
Had to revert this in https://github.com/Homebrew/formulae.brew.sh/pull/118 unfortunately.

.gitignore
```diff
@@ -1,3 +1,4 @@
+.DS_Store
```
MikeMcQuaid: @EricFromCanada When splitting this out into a new PR: could you make a separate one just for the non-Cask-specific changes like this (and the formula changes) so I can merge that immediately? Thanks!

_layouts/analytics.html
```diff
@@ -20,7 +20,9 @@ <h3><a href="{{ site.baseurl }}/api/analytics/{{ page.category }}/{{ page.days}}
     {%- if page.category == "os-version" -%}
       <code>{{ item.os_version }}</code>
     {%- elsif page.category == "cask-install" -%}
-      <code>{{ item.cask }}</code>
+      <a href="{{ site.baseurl }}/cask/{{ item.cask }}">
```
MikeMcQuaid: This needs to check if the cask actually exists (e.g. versioned casks do not) like is done with formulae below.

index.html
```diff
@@ -5,5 +5,6 @@
 and use Homebrew see <a href="https://brew.sh">our homepage</a>.</p>
 
 <h2><a href="{{ site.baseurl }}/formula/">Browse all formulae</a></h2>
+<h2><a href="{{ site.baseurl }}/cask/">Browse all Casks</a></h2>
```
MikeMcQuaid: Lowercase the `Casks` to match `formulae`.

cask_index.html
```diff
@@ -0,0 +1,19 @@
...
+<p>This is a listing of all Casks available via the <a href="https://brew.sh">Homebrew</a> package manager for macOS.</p>
+
+<h2><a href="{{ site.baseurl }}/api/cask.json"><code>/api/cask.json</code> (JSON API)</a></h2>
+
+<table>
+    {%- assign sorted_casks = site.data.cask | sort -%}
+    {%- for cask in sorted_casks -%}
+
+    <tr>
+        {%- assign t = cask[0] -%}
```
MikeMcQuaid: `assign token`

cask_index.html
```diff
@@ -0,0 +1,19 @@
...
+        {%- assign c = cask[1] -%}
```
MikeMcQuaid: `assign cask`

cask_index.html
```diff
@@ -0,0 +1,19 @@
...
+        {%- include cask.html token=t cask=c -%}
```
MikeMcQuaid: Is it worth including both the Cask token and the Cask name? They are very similar. May be worth having the version follow the name like for formulae.

.gitignore
EricFromCanada: Submitted as #121.

cask_index.html
EricFromCanada: The cask token has to be passed to `_includes/cask.html` because, unlike formulae, it's not actually included within the `cask[1]` data structure.

MikeMcQuaid: Sorry, to be explicit: I mean is it worth outputting them both on the index page in separate columns?

EricFromCanada: 1. search engines: since cask tokens don't always match their names letter-for-letter, including both strengthens the association for when search engines crawl the site.
2. cask discoverability: when viewing the list of casks, including the cask name allows a user doing an in-page search of, say, "PyCharm Community Edition" to jump to the `pycharm-ce` row.

MikeMcQuaid: Works for me 👍

PR #118, opened 2019-03-23
Revert "add support for Casks"

MikeMcQuaid:
Reverts Homebrew/formulae.brew.sh#114 which was merged prematurely because the GitHub Actions haven't been fixed to work on Linux yet: https://github.com/Homebrew/homebrew-cask/runs/83474463

Sorry @EricFromCanada. @reitermarkus, @cmbernard333 or @EricFromCanada: any interest in looking into this?


cmbernard333:
I could probably look into this. How do we reproduce it locally?

MikeMcQuaid:
The short answer is "run it on Linux". Otherwise you may be able to by setting `HOMEBREW_TEST_GENERIC_OS` or writing a test to reproduce and running `brew tests --generic`.

EricFromCanada:
With the change above plus those from #114, `rake cask` completes on Ubuntu 16.04.

PR #126, opened 2019-03-27
fix shadowed formula names

EricFromCanada:
The `site.data.formula` data structure omits plus signs from its keys. This is a problem for a few pairs of formulae that differ only by plus signs; specifically `mysql`/`mysql++`, `mysql-connector-c`/`mysql-connector-c++`, `pcre`/`pcre++`, causing only one to appear in the formula list and [detail](https://formulae.brew.sh/formula/pcre) [pages](https://formulae.brew.sh/formula/pcre++). This change replaces plus signs with underscores when generating `_data/formula/*.json` files so that both appear.


MikeMcQuaid:
Good catch, thanks again @EricFromCanada!

PR #128, opened 2019-03-28
use double quotes around strings in liquid tags

EricFromCanada:
… and update one file missed in #126.


MikeMcQuaid:
Oops, good catch, thanks again @EricFromCanada!

PR #164, opened 2019-08-20
Update TomEE Webprofile version to latest (7.1.1)

nfrankel:
This PR updates the version of the TomEE Webprofile to the latest. The version currently referenced is not available anymore.


EricFromCanada:
You need to issue a PR on this file instead: https://github.com/Homebrew/homebrew-core/blob/master/Formula/tomee-webprofile.rb

nfrankel:
Thanks @EricFromCanada for the pointer

PR #165, opened 2019-08-24
Update FreeCAD sha256 checksum

NickoJ:
sha256 checksum was incorrect


MikeMcQuaid:
Submit this to homebrew-core.

PR #168, opened 2019-08-30
Suppress control characters around text using Formatter

EricFromCanada:
e.g. caveats for 'cockroach' formula or 'adoptopenjdk' cask. Only seems to be an issue when building locally, though.

cc @MikeMcQuaid


MikeMcQuaid:

PR #191, opened 2019-10-14
CI: Switch Actions 'version' parameter for 'ruby-version'

issyl0:
- The Actions UI showed a deprecation warning for `version`:
  Input 'version' has been deprecated with message: The version property will not be supported after October 1, 2019. Use ruby-version instead

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Set up Ruby
       uses: actions/setup-ruby@v1
       with:
-        version: '>=2.3'
+        ruby-version: '>=2.3'
 
     - name: Install RubyGems
       run: |


.github/workflows/tests.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - name: Set up Ruby
       uses: actions/setup-ruby@v1
       with:
-        version: '>=2.3'
+        ruby-version: '>=2.3'
```
issyl0: Should we also bump this to 2.6 given the recent upgrade of Homebrew/brew?

```suggestion
        ruby-version: '>=2.6'
```

MikeMcQuaid:
Thanks @issyl0!

.github/workflows/tests.yml
MikeMcQuaid: Whoops, just missed this! Yes, this seems like a good idea to do across all the repos in the org?

issyl0: Cool, I’ll do this tomorrow unless anyone gets there first.

PR #192, opened 2019-10-17
Use double quotes

isnullxbh:
We cannot use command substitution in expressions wrapped into single quotes; expression will output as is;
So, we expect that execution of the following command:
`'$(brew --prefix)/share'`
should be:
 '/usr/local/share' (as example)
, but we get:
'$(brew --prefix)/share'

### DIFF
--- a/_data/formula/zsh-completions.json
+++ b/_data/formula/zsh-completions.json
@@ -46,7 +46,7 @@
   "conflicts_with": [
 
   ],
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Submit this to Homebrew/homebrew-core.

isnullxbh:
> Submit this to Homebrew/homebrew-core.

Hello, @MikeMcQuaid!
Explain, please. I would like to fix error, that I found on this page:
https://formulae.brew.sh/formula/zsh-completions

I'm talking about following command:
chmod go-w '$(brew --prefix)/share'

Please, explain how and where I can fix it?

MikeMcQuaid:
Submit a PR to https://github.com/homebrew/homebrew-core/.

isnullxbh:
> Submit a PR to https://github.com/homebrew/homebrew-core/.

I made pull-request for mentioned repository, but it seems I was fix wrong file.
https://github.com/Homebrew/homebrew-core/pull/45380

PR #193, opened 2019-10-21
Use the formula_path variable to get the OS-scoped formula name

issyl0:
- This was hardcoded to the macOS formula data directory by mistake, causing everything to work fine _except_ Linux-only formulae which obviously aren't present in the macOS data files.
- The analytics part of this file was using the `formula_path` variable correctly, which meant that the data appeared fine, but the Linux-only formulae just didn't have names or anything that relied on a lookup based on name.

Before, `/formula-linux/adoptopenjdk`:

image

After:

image


MikeMcQuaid:
Nice work!

PR #196, opened 2019-10-28
Really fix dependency list path for Linux formulae

issyl0:
- #195 was almost there but not quite!
- We need to use `formula_path` in two places: once to get the correct formula data, and once to specify the path for the formula.html includes - otherwise, as was happening here, it defaults to `formula`.


MikeMcQuaid:
Thanks again @issyl0!

PR #266, opened 2020-04-04
Make tap configurable

jasonkarns:
This PR is an attempt to make the tap configurable, such that forks may build the website for their own 3rd party taps.

This PR is comprised of two primary changes:

- Rakefile and generator scripts:
`script/generate.rb` and `script/generate-cask.rb` are modified to accept a tap argument of the form "owner/repo". If the argument is omitted or empty, they default to their present values (CoreTap and default_cask_tap, respectively). To pair with this change, the corresponding Rakefile tasks are also modified to accept a tap argument and to pass it along to the scripts.

- Tap info in Jekyll configuration
The second main change in this PR, is the addition of tap information to the Jekyll configuration. These configuration options are then used in the cask and formula templates when constructing URLs for the formulae. Again, the configuration is pre-set to the homebrew-core and cask values.

With both of these changes in place, it becomes easier for the site to be built for 3rd party taps, but still defaults to the primary core/cask taps.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -15,10 +15,12 @@ task :formulae, [:os,:tap] do |task, args|
 end
 
 desc "Dump cask data"
-task :cask do
+task :cask, [:tap] do |task, args|
+  args.with_defaults(:tap => "")
+
   ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1"
   ENV["HOMEBREW_NO_COLOR"] = "1"
-  sh "brew", "ruby", "script/generate-cask.rb"
+  sh "brew", "ruby", "script/generate-cask.rb", args[:tap]
 end
 
 def generate_analytics?(os)
--- a/Rakefile
+++ b/Rakefile
@@ -6,12 +6,12 @@ require "date"
 task default: :formula_and_analytics
 
 desc "Dump macOS formulae data"
-task :formulae, [:os] do |task, args|
-  args.with_defaults(:os => "mac")
+task :formulae, [:os,:tap] do |task, args|
+  args.with_defaults(:os => "mac", :tap => "")
 
-  ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1" if args.os == "mac"
+  ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1" if args[:os] == "mac"
   ENV["HOMEBREW_NO_COLOR"] = "1"
-  sh "brew", "ruby", "script/generate.rb", args.os
+  sh "brew", "ruby", "script/generate.rb", args[:os], args[:tap]
 end
 
 desc "Dump cask data"
--- a/Rakefile
+++ b/Rakefile
@@ -6,7 +6,7 @@ require "date"
 task default: :formula_and_analytics
 
 desc "Dump macOS formulae data"
-task :formulae, [:os,:tap] do |task, args|
+task :formulae, [:os, :tap] do |task, args|
   args.with_defaults(:os => "mac", :tap => "")
 
   ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1" if args[:os] == "mac"
--- a/Rakefile
+++ b/Rakefile
@@ -7,7 +7,7 @@ task default: :formula_and_analytics
 
 desc "Dump macOS formulae data"
 task :formulae, [:os, :tap] do |task, args|
-  args.with_defaults(:os => "mac", :tap => "")
+  args.with_defaults(:os => "mac", :tap => "homebrew/core")
 
   ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1" if args[:os] == "mac"
   ENV["HOMEBREW_NO_COLOR"] = "1"
@@ -16,7 +16,7 @@ end
 
 desc "Dump cask data"
 task :cask, [:tap] do |task, args|
-  args.with_defaults(:tap => "")
+  args.with_defaults(:tap => "homebrew/cask")
 
   ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1"
   ENV["HOMEBREW_NO_COLOR"] = "1"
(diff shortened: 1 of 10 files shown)


Rakefile
```diff
@@ -6,19 +6,21 @@ require "date"
 task default: :formula_and_analytics
 
 desc "Dump macOS formulae data"
-task :formulae, [:os] do |task, args|
-  args.with_defaults(:os => "mac")
+task :formulae, [:os,:tap] do |task, args|
+  args.with_defaults(:os => "mac", :tap => "")
 
-  ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1" if args.os == "mac"
+  ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1" if args[:os] == "mac"
```
jasonkarns, line 12: Had to switch `args.tap` to `args[:tap]` because .tap conflicts with the `Object.tap` ruby method. To keep things consistent within this block, the .os invocations were converted as well.

Rakefile
```diff
@@ -6,19 +6,21 @@ require "date"
 task default: :formula_and_analytics
 
 desc "Dump macOS formulae data"
-task :formulae, [:os] do |task, args|
-  args.with_defaults(:os => "mac")
+task :formulae, [:os,:tap] do |task, args|
```
MikeMcQuaid, line 9:
```suggestion
task :formulae, [:os, :tap] do |task, args|
```

script/generate-cask.rb
```diff
@@ -1,13 +1,17 @@
 #!/usr/bin/env brew ruby
 require "cask/cask"
+
+tap_name = (ARGV.first || "").split("/")
+tap = tap_name.empty? ? Tap.default_cask_tap : Tap.new(*tap_name)
```
MikeMcQuaid, line 5:
```suggestion
tap_name = ARGV.first
tap = Tap.new(tap_name.split("/")) if tap_name
tap ||= Tap.default_cask_tap
```

script/generate.rb
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env brew ruby
 os = ARGV.first
+tap_name = (ARGV.second || "").split("/")
 
 formula_dir = os == "mac" ? "formula" : "formula-linux"
+tap = tap_name.empty? ? CoreTap.instance : Tap.new(*tap_name)
```
MikeMcQuaid, line 6:
```suggestion
tap_name = ARGV.second
tap = Tap.new(tap_name.split("/")) if tap_name
tap ||= CoreTap.instance

formula_dir = os == "mac" ? "formula" : "formula-linux"
```

jasonkarns: @MikeMcQuaid Question: is it preferred for the default core tap to be specified as `CoreTap.instance`? ie, is there some machinery within homebrew that would make `CoreTap.instance` preferable to `Tap.new("homebrew/core")`? If not, we could simplify all this logic by having just a single code path which defaults the tapname argument to `"homebrew/core"`.

jasonkarns: Actually I just noticed the `Tap.fetch` class method, which seems to do all the normalization, caching and singleton-preference all at once. I think I'll switch to that instead of `Tap.new` ?

jasonkarns:
> I wonder if there's any pages like API/analytics etc. that are worth hiding on tap stuff by default?

That's a good question!

As you can guess this is related to the discussion on discourse regarding analytics data for 3rd part taps. Taps don't have the ability to generate our _own_ JSON files from GA using the formula-analytics subcommand (due to the GA access restriction). However, as I learned from that thread, tap-specific data is _already present_ in these JSON files on formula.brew.sh. So what I did for my tap's fork of the site was to replace the generate_analytics functions in the Rakefile to `curl` the corresponding JSON files directly from formula.brew.sh, and run them through a `jq` filter to select out only the formula that live in my tap.

https://github.com/nodenv/formulae/blob/ae6bc0787539a72d91436f2dc695bcd7de5e23b4/Rakefile#L32-L58

I can't think of a clean way to have logic like this exist in the main site without being super kludgy. (And let's be honest, there aren't going to be many tap-specific forks of this site.) I thought the changes in this PR were simple enough to be worth it; but the kludginess of the analytics stuff is too costly to warrant living in upstream (IMO).

So that accounts for the main event categories (install, install-on-request, cask-install, and build-errors). The rest are already specific to the homebrew-core or linuxbrew, or don't really apply to taps (the macOS version data). So for my tap's fork I just commented them out completely.

I have yet to go through the JSON API documentation page but I assume that one will remain mostly untouched.


The only big remaining piece is the algolia search. I've reached out to algolia to see about setting up a separate account/index for our fork. The only issue is that the algolia configuration lives in the main brew.sh repo. I'm going to experiment with ways to make that configurable, but that's probably beyond the scope of this PR for now.

script/generate.rb
MikeMcQuaid: Sounds good. Happy to use `"homebrew/core"` over `CoreTap.instance`, too.

MikeMcQuaid:
> I can't think of a clean way to have logic like this exist in the main site without being super kludgy. (And let's be honest, there aren't going to be many tap-specific forks of this site.) I thought the changes in this PR were simple enough to be worth it; but the kludginess of the analytics stuff is too costly to warrant living in upstream (IMO).

Fine with me 👍. Certainly doesn't need to be in this PR.

Shout when you're good to merge (from your perspective)!

Also, if you can be bothered, even some really basic/crappy docs on how to setup a fork of this would be neat.

Thanks again!

jasonkarns:
Latest commit updates core tap references from the api docs page.

The `Homebrew/homebrew-core` references will appear unchanged. However, the `homebrew-core` references will now appear as `core`. Prior to addressing the API docs, all other tap references were either shortname (homebrew/core) or fullname (Homebrew/homebrew-core), so the current yaml config doesn't provide a key for the "full repo name".

If this change is acceptable, I consider this PR ready to merge at this point. I'll work on a separate PR that handles the {{ site.baseurl }} on the API docs page, and includes some docs for tap-specific forks.

MikeMcQuaid:
Thanks so much for your contribution! Without people like you submitting PRs we couldn't run this project. You rock, @jasonkarns!

MikeMcQuaid:
> If this change is acceptable, I consider this PR ready to merge at this point. I'll work on a separate PR that handles the {{ site.baseurl }} on the API docs page, and includes some docs for tap-specific forks.

Looking forward to it!

MikeMcQuaid:
https://nodenv.github.io/formulae/ works surprisingly well, nice work! Makes me even more glad I decided to ~hack~ build this on GitHub Pages!

jasonkarns:
@MikeMcQuaid For sure, Mike! I _maybe_ would still have attempted a fork if it were on heroku, but any other amount of work, I seriously doubt it. GitHub Pages is perfect.

Thanks for all your work!

Issue #267, opened 2020-04-04
Rake task for linux is skipped when running build


jasonkarns:
# Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.

- [x] your issue is with the https://formulae.brew.sh website and not the (generated) contents of a given formula/cask page.

## What you were trying to do (and why)

I am trying to ensure both mac and linux rake tasks are run prior to build, to ensure _all_ the data is generated successfully.

## What happened (include screenshots)

When running `rake build`, it runs the rake task for generating mac formulae/analytics successfully, and then it runs the cask task. However, when it runs the linux task (`linux_formula_and_analytics`), the `formulae` and `analytics` are skipped because rake considers the tasks already run.

<img width="924" alt="image" src="https://user-images.githubusercontent.com/119972/78456994-1f80b000-7675-11ea-8bde-edb3726432cb.png">

Note that `script/generate.rb` did _not_ run for linux, despite the `linux_formula_and_analytics` task being a pre-req for `build`.

(And yes, the tasks here have an extra argument, the tap, which is from PR #266. But that is the only change to my rakefile.)

## What you expected to happen

Personally, I expected build to run all three main tasks (formulae+analytics for mac, formulae+analytics for linux, and the cask task). However, it's possible that the build task is intentionally leveraging rake's behavior to only run _either_ mac or linux tasks, and not both. But if that's the intention, it wasn't clear to me. (It's such a subtle rake 'gotcha' that I wouldn't be surprised if this is unexpected or intentional.)

If it is actually expected that both mac and linux tasks run for `build`, then I'd be happy to open a PR to make the linux task `reenable` both the formulae and analytics tasks prior to invoking them.

jasonkarns:
Apparently, the `test` workflow in this repo works around this limitation by manually invoking all three tasks from the command line: https://github.com/Homebrew/formulae.brew.sh/blob/ac27213f993a5efa9ac6da48dd339a876eb07e21/.github/workflows/tests.yml#L41

So there is clearly a desire to be able to run all three (mac, linux, cask), but it presently can't be done with a single task.

MikeMcQuaid:
> If it is actually expected that both mac and linux tasks run for `build`, then I'd be happy to open a PR to make the linux task `reenable` both the formulae and analytics tasks prior to invoking them.

[...]

PR #269, opened 2020-04-06
Ensure linux rake tasks run

jasonkarns:
As discussed in #267, the Linux formulae and analytics task(s) are skipped whenever they are run after the default (macOS) tasks. This is because rake considers the tasks already satisfied (the manual invocation still checks whether the task has already been run).

This PR makes the Linux task manually `reenable` the formula and analytics tasks prior to invoking them.

In addition to making `linux_formula_and_analytics` run as desired (which ensures the `build` task runs as expected), this PR also introduces `all_formulae` and `all_analytics` tasks, which ensure both macOS and Linux variants of the corresponding tasks are run. (Better names for these tasks would be welcome.)

The tests workflow is then updated to invoke these new tasks.

Lastly, this PR also makes the symbol array usage consistent within the Rakefile (matching the existing use of `%i[]`). And updates the jekyll task to invoke jekyll programmatically (ala the html proofer and jsonlint tasks) in order to avoid shelling-out.


jasonkarns:
Looks like the request-info bot doesn't ignore Draft PRs 😄  (I had opened as draft to get the build running before I updated the PR title/description. It has all been updated, but presumably the needs-response label doesn't go away automatically? 🤷 Regardless, the title and description have been updated 😄

MikeMcQuaid:
Makes sense, thanks again @jasonkarns! Will merge once conflicts resolved.

jasonkarns:
all set!

PR #271, opened 2020-04-16
Rake updates

jasonkarns:
The primary purpose of this PR is to update the Rakefile such that the data generation tasks pull their default tap names from the jekyll configuration. The benefit of this is reducing the number of places where this information is duplicated (thereby simplifying the configuration changes necessary for tap-forks).

Additionally, this PR configures the rake CLOBBER listing such that the `rake clobber` task is capable of clearing out not only jekyll's generated site, but also the generated formula/cask/analytics data files. Rake differentiates clean and clobber where clobber is a more extensive or "thorough" cleaning. I think this makes a very good differentiating between cleaning the jekyll site (which is not committed) and clobbering the homebrew data (which _is_ committed). Clobber is a much less rarely used task, after all.

Accordingly, I have added the clean/clobber modifications next to the corresponding rake tasks which generate the respective files and directories. I believe this pattern ensures that the clean/clobber lists will be maintained over time as the rake tasks themselves are maintained; ensuring they stay in sync.

Lastly, this adds a `serve` rake task to correspond to the jekyll-serve command. I find it useful to treat rake as the *primary* source for tasks. `serve` was the only "primary" command that was still left to be run outside of rake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ To instead run Homebrew Formulae locally, run:
 git clone https://github.com/Homebrew/formulae.brew.sh
 cd formulae.brew.sh
 bundle install
-bundle exec jekyll serve
+bundle exec rake serve
 ```
 
 ## License
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ To instead run Homebrew Formulae locally, run:
 git clone https://github.com/Homebrew/formulae.brew.sh
 cd formulae.brew.sh
 bundle install
-bundle exec rake serve
+bundle exec jekyll serve
 ```
 
 ## License
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,6 @@
 require "rake"
 require "rake/clean"
+require 'jekyll'
 require "json"
 require "date"
 
@@ -7,7 +8,7 @@ task default: :formula_and_analytics
 
 desc "Dump macOS formulae data"
 task :formulae, [:os, :tap] do |task, args|
-  args.with_defaults(:os => "mac", :tap => "homebrew/core")
+  args.with_defaults(:os => "mac", :tap => Jekyll.configuration.dig("taps", "core", "name"))
 
   ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1" if args[:os] == "mac"
   ENV["HOMEBREW_NO_COLOR"] = "1"
@@ -16,7 +17,7 @@ end
 
 desc "Dump cask data"
 task :cask, [:tap] do |task, args|
-  args.with_defaults(:tap => "homebrew/cask")
+  args.with_defaults(:tap => Jekyll.configuration.dig("taps", "cask", "name"))
 
   ENV["HOMEBREW_FORCE_HOMEBREW_ON_LINUX"] = "1"
   ENV["HOMEBREW_NO_COLOR"] = "1"
--- a/Rakefile
+++ b/Rakefile
@@ -14,6 +14,8 @@ task :formulae, [:os, :tap] do |task, args|
   ENV["HOMEBREW_NO_COLOR"] = "1"
   sh "brew", "ruby", "script/generate.rb", args[:os], args[:tap]
 end
+CLOBBER.include FileList[%w[_data/formula api/formula formula
+                         _data/formula-linux api/formula-linux formula-linux]]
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -24,7 +24,7 @@ To instead run Homebrew Formulae locally, run:
 git clone https://github.com/Homebrew/formulae.brew.sh
 cd formulae.brew.sh
 bundle install
-bundle exec jekyll serve
+bundle exec rake serve
```
MikeMcQuaid, line 27: Would rather leave this as-is

Rakefile
```diff
@@ -124,7 +129,13 @@ end
 
 desc "Build the site"
 task build: [:formula_and_analytics, :cask, :linux_formula_and_analytics] do
-  sh "bundle", "exec", "jekyll", "build"
+  Jekyll::Commands::Build.process({})
+end
+CLEAN.include FileList["_site"]
+
+desc "Serve the site (jekyll serve)"
```
MikeMcQuaid, line 136:
```suggestion
desc "Serve the site"
```

jasonkarns:
@MikeMcQuaid Applied suggestions!

MikeMcQuaid:
Thanks again @jasonkarns!

Bo98:
Posting here as someone who's never touched `rake` or `jekyll`:

[This workflow](https://github.com/Homebrew/homebrew-core/blob/master/.github/workflows/generate_formulae.brew.sh_data.yml#L33) seems to no longer work:

```
rake aborted!
LoadError: cannot load such file -- jekyll
/Users/runner/runners/2.169.0/work/homebrew-core/homebrew-core/formulae.brew.sh/Rakefile:3:in `<top (required)>'
-e:1:in `load'
-e:1:in `<main>'
```

The workflow might need changing, not sure.

MikeMcQuaid:
@jasonkarns Mind taking a look at this?

jasonkarns:
Oof! I didn't realize the rake tasks in this repo were invoked from "outside" the repo. The problem appears to be, because some of these rake tasks are invoked without this repo's gems (jekyll in particular) being installed. Options:

1. We can use stdlib `yaml` to load _config.yml instead of using jekyll itself to expose its configuration.
2. We can have these other repos bundle-install before invoking their tasks.
3. We can have the other repos send repository-dispatch events to this repo to trigger the data generation tasks, instead of running those tasks "externally".

No. 1 is the easiest, and I'll open a PR to do that right away to fix the problem. However, I don't think it's the "right" solution long term. I believe the problem is due to the boundaries of these repo being crossed. Other repos are invoking this repo's tasks without a complete ruby/bundle environment; and they are doing so in order to commit changes into this repo. That feels like a boundary violation. I think No. 3 would resolve this problem more cleanly and appropriately. It keeps this repo in control of the workflows which mutate its data, and keeps other repos' "hands" out of its rake environment. But it'll require a bit more work, hence No.1 being a good short term solution. (And as a bonus, I believe it will simplify all the other repos' workflows, since all the data generation workflows will be consolidated here.)

Bo98:
Number 2 seems trivial to do: https://github.com/Homebrew/homebrew-core/pull/53300, though maybe you have a point about number 3.

jasonkarns:
@Bo98 An additional problem with No2 is that it would need to be replicated across all of the taps that use the formulae generation workflows. Which (I believe) are homebrew-core, linuxbrew-core, and homebrew-cask (maybe even more?). I'm not excited about duplicating that change across multiple repos; especially since the change that necessitates it is isolated to just here. (Though, again, it's a boundary/context violation, so they probably _should_ have been bundle-installing from the beginning? 🤷 )

I do have a branch up to do No.1. I'm testing it with my own fork of homebrew-core, before I PR it.

jasonkarns:
@Bo98 PR for No 1 [open](https://github.com/Homebrew/formulae.brew.sh/pull/272). Apologies for breaking the workflows 😬

MikeMcQuaid:
> 3\. We can have the other repos send repository-dispatch events to this repo to trigger the data generation tasks, instead of running those tasks "externally".

I think this is a nice option 👍. Another that has been proposed is just making it a scheduled Actions workflow on this repo that could checkout out the other repos if a e.g. stored hash in this repo isn't the tip of that repo's `master` (without needing a clone you can do this with https://developer.github.com/changes/2016-02-24-commit-reference-sha-api/).

PR #272, opened 2020-04-18
Allow use of rake without a gem bundle

jasonkarns:
This allows the data generation tasks to be run outside the context of a
Bundler gem bundle. This is necessary for the core and cask taps to
trigger data re-generation from their Action workflows.

Hopefully this is a short-term solution, as it's preferable for rake to
be able to assume (and leverage) access to its bundled gems.

see https://github.com/Homebrew/formulae.brew.sh/pull/271#issuecomment-615897164

[Example invocation](https://github.com/jasonkarns/homebrew-core/runs/598148507?check_suite_focus=true) of rake from an external tap (my fork of homebrew-core, pointing at this PR's branch of formulae.brew.sh)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,14 +1,19 @@
+require "date"
+require "json"
 require "rake"
 require "rake/clean"
-require 'jekyll'
-require "json"
-require "date"
+require "yaml"
+
+def jekyll_config(*props)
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Great work, thanks for quick turnaround here @jasonkarns!

PR #275, opened 2020-04-20
Bump Ruby to 2.6.6

issyl0:
- This was quite far behind the latest at 2.6.2, and the
  `actions/setup-ruby` action has been running 2.6.6 only for some days
  now, so it must work.


.ruby-version
```diff
@@ -1 +1 @@
-2.6.2
+2.6.6
```
MikeMcQuaid, line 1:
```suggestion
2.6
```

and then it shouldn't need bumped again for a while 😀

issyl0: Fine, there you go!

MikeMcQuaid:
Thanks @issyl0!

PR #277, opened 2020-05-02
actions: periodically generate site data

EricFromCanada:
[As suggested](https://github.com/Homebrew/homebrew-core/pull/53939#issuecomment-620959807), assume the task of periodically regenerating site data from changes in homebrew-core, homebrew-cask, and linuxbrew-core. Will need `secrets.ANALYTICS_JSON_KEY` and `secrets.ANALYTICS_JSON_KEY_B64` added to this repo for analytics processing to work.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -1,0 +1,123 @@
+name: Scheduled regeneration
+
+on:
+  schedule:
+    # Once every other hour
+    - cron:  '55 */2 * * * *'
+
+jobs:
+  generate:
+    if: startsWith( github.repository, 'Homebrew/' )
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macOS-latest, ubuntu-latest]
+    env:
+      CI: 1
+      HOMEBREW_COLOR: 1
+      HOMEBREW_NO_ANALYTICS: 1
+      HOMEBREW_NO_AUTO_UPDATE: 1
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+
+      - name: Set up Ruby
+        if: matrix.os == 'ubuntu-latest'
+        uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.6'
+
+      - name: Set up Homebrew
+        run: |
+          if which brew &>/dev/null; then
+            HOMEBREW_REPOSITORY="$(brew --repository)"
+            brew update-reset "$HOMEBREW_REPOSITORY"
+          else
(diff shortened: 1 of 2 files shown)


issyl0:
All the memories are coming back now. 😄

.github/workflows/scheduled.yml
```diff
@@ -0,0 +1,112 @@
...
+          git add _data/cask api/cask cask
+
+          if ! git diff --no-patch --exit-code HEAD -- _data/cask api/cask cask; then
+            git commit -m "cask: updates from homebrew/cask" _data/cask api/cask cask
+            git fetch
+            git rebase origin/master
+            git push
+          fi
+
+  generate-linux:
```
MikeMcQuaid, line 66: Thoughts on making this a matrix job so it can share more code?

MikeMcQuaid: (check the homebrew-core or homebrew-test-bot repos for prior art for e.g. the setup)

EricFromCanada: Any ideas for a container to apply to only one OS in a matrix? [No luck](https://github.com/EricFromCanada/formulae.brew.sh/actions/runs/94490949) with that so far…

MikeMcQuaid: No ideas sorry 😭

MikeMcQuaid:
Looks great, merge when ready @EricFromCanada. I think it might be nice to add a "generated at" timestamp or something to the page footer?

EricFromCanada:
One odd behaviour I'm seeing is updates for Homebrew/linuxbrew-core data [translating requirements into Casks](https://github.com/EricFromCanada/formulae.brew.sh/commit/3e5552d67cb0236baaa0cd29908b01de9d7676e5), which should be a Mac-only behaviour.

EricFromCanada:
> My guess: openjdk is installed on the Actions node through the system package manager, and abcl uses depends_on :java , which picks up the Requirement from the system instead of fulfilling it via a Homebrew formula. (@jonchang)

Will leave the question of whether a Linux formula's JSON output should be referencing casks for another time.

EricFromCanada:
> add a "generated at" timestamp or something to the page footer?

@MikeMcQuaid By this do you mean the footer of each individual formula's web page, or just other select pages?

MikeMcQuaid:
> @MikeMcQuaid By this do you mean the footer of each individual formula's web page, or just other select pages?

I think even just a site-wide footer on the homepage (and in JSON files) would be useful.

jonchang:
Nice work on this!

MikeMcQuaid:
Thanks @EricFromCanada and for cleaning up all the existing jobs too 😻

PR #280, opened 2020-05-06
Generate initial homebrew/cask analytics

EricFromCanada:
And allow them to be generated by not setting `HOMEBREW_NO_ANALYTICS` globally.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -13,7 +13,6 @@ jobs:
     env:
       CI: 1
       HOMEBREW_COLOR: 1
-      HOMEBREW_NO_ANALYTICS: 1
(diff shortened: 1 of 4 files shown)


.github/workflows/scheduled.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     env:
       CI: 1
       HOMEBREW_COLOR: 1
-      HOMEBREW_NO_ANALYTICS: 1
+      HOMEBREW_NO_ANALYTICS_THIS_RUN: 1
```
MikeMcQuaid, line 16: This isn't meant to be a user-accessible variable (it's not documented). Could `HOMEBREW_NO_ANALYTICS` be set/unset more specifically where desired?

PR #289, opened 2020-05-16
Improve display of macOS requirements for casks

EricFromCanada:
Properly handle multiple ranges: `macOS >= 10.11 and macOS <= 10.14` and multiple versions: `macOS == 10.11 / 10.12 / 10.13 / 10.14 / 10.15`.


MikeMcQuaid:
Thanks again @EricFromCanada!

PR #292, opened 2020-05-19
Always show analytics table on formula/cask pages

EricFromCanada:
Original logic would skip showing the table if the formula hadn't had any installs in the last 30 days. This also increases the DRY factor by looping through each days interval.


MikeMcQuaid:

PR #317, opened 2020-08-18
Show additional formula/cask info

EricFromCanada:
- sets a "disabled" or "deprecated" class and hovertext for titles of such formulae (relies on changes to site theme in https://github.com/Homebrew/brew.sh/pull/516) in both the formula page title and formula listings
- displays cask descriptions if they exist
- replaces the underscore with a space for formulae specifying `license :public_domain`

As suggested in https://github.com/Homebrew/brew/pull/8166.


MikeMcQuaid:
Very nice, thanks again @EricFromCanada!

PR #330, opened 2020-09-21
Add install command and copy to clipboard

### LINKED ISSUE #329: Idea: Copy install command to clipboard button
- [x] your issue is with the https://formulae.brew.sh website and not the (generated) contents of a given formula/cask page.

## What you were trying to do (and why)

install a cask

## What happened (include screenshots)

I wasted some time waiting for brew update to finish and tell me that it was a cask and not a formula

## What you expected to happen

A button which copies the install command to clipboard, so all a user has to do is Cmd + V

This was discussed in #270

sidharthv96:
Adds install command user can directly copy and paste.

Fixes #329 
image

Points to finalise
- [x] Placement of element
- [x] Icon to use

### DIFF
--- a/_includes/code_snippet.html
+++ b/_includes/code_snippet.html
@@ -1,0 +1,17 @@
+{% assign code = include.code %}
+<style>
+.copyButton {
+	background: transparent;
+    border: none;
+}
+</style>
+<script src="https://cdn.jsdelivr.net/npm/clipboard@2.0.6/dist/clipboard.min.js"></script>
+
+<code id="code">{{code}}</code>
+<button class="copyButton" data-clipboard-target="#code" aria-label="Copy to clipboard">
+    📋
+</button>
+
+<script>
+    new ClipboardJS('.copyButton');
+</script>
--- a/_includes/code_snippet.html
+++ b/_includes/code_snippet.html
@@ -3,6 +3,7 @@
     #copyButton {
         background: transparent;
         border: none;
+        visibility: hidden;
     }
 </style>
 
@@ -16,8 +17,9 @@
 </button>
 
 <script>
-    if(!navigator.clipboard) {
-        document.getElementById("copyButton").remove();
+    if (navigator.clipboard) {
+        document
+            .getElementById("copyButton")
+            .style.removeProperty("visibility");
     }
 </script>
-
--- a/_includes/code_snippet.html
+++ b/_includes/code_snippet.html
@@ -15,3 +15,9 @@
     📋
 </button>
 
+<script>
+    if(!navigator.clipboard) {
+        document.getElementById("copyButton").remove();
+    }
+</script>
+
--- a/_includes/code_snippet.html
+++ b/_includes/code_snippet.html
@@ -1,17 +1,20 @@
 {% assign code = include.code %}
 <style>
-.copyButton {
-	background: transparent;
-    border: none;
-}
+    .copyButton {
+        background: transparent;
+        border: none;
+    }
 </style>
-<script src="https://cdn.jsdelivr.net/npm/clipboard@2.0.6/dist/clipboard.min.js"></script>
 
-<code id="code">{{code}}</code>
-<button class="copyButton" data-clipboard-target="#code" aria-label="Copy to clipboard">
+<code>{{code}}</code>
+<button
+    id="copyButton"
+    aria-label="Copy to clipboard"
+>
     📋
 </button>
 
 <script>
-    new ClipboardJS('.copyButton');
+    document.querySelector("#copyButton")
+    .addEventListener("click", () => navigator.clipboard.writeText("{{code}}"));
 </script>
--- a/_includes/code_snippet.html
+++ b/_includes/code_snippet.html
@@ -10,11 +10,8 @@
 <button
     id="copyButton"
     aria-label="Copy to clipboard"
+    onclick="navigator.clipboard.writeText('{{code}}')"
 >
     📋
 </button>
 
-<script>
-    document.querySelector("#copyButton")
-    .addEventListener("click", () => navigator.clipboard.writeText("{{code}}"));
-</script>
--- a/_includes/code_snippet.html
+++ b/_includes/code_snippet.html
@@ -1,6 +1,6 @@
 {% assign code = include.code %}
 <style>
-    .copyButton {
+    #copyButton {
         background: transparent;
         border: none;
     }
--- a/_includes/install_command.html
+++ b/_includes/install_command.html
@@ -14,9 +14,11 @@ brew {{include.modifier}} install {{include.item}}
 </style>
 
 Install command:
-<pre class="inline">
-    <code>{{code}}</code>
-</pre>
+<div class="inline">
+    {% highlight bash %}
+    $ {{code}}
+    {% endhighlight %}
+</div>
 <button
     id="copyButton"
     aria-label="Copy to clipboard"
--- a/_includes/install_command.html
+++ b/_includes/install_command.html
@@ -1,49 +1,10 @@
-{%- capture code -%}
-brew {{include.modifier}} install {{include.item}}
-{%- endcapture -%}
-<style>
-    .inline {
-        display: inline-flex;
-        background: rgba(0, 0, 0, 0.3);
-        border-radius: 0.4em;
-        padding-right: 8px;
-    }
-
-    .inline .highlight {
-        margin: 0;
-    }
-    
-    .inline .highlight pre {
-        margin: 0;
-    }
-    
-    #copyButton {
-        background: transparent;
-        border: none;
-        border-left: 1px solid #aaa;
-    }
-</style>
+{%- capture code -%} 
+brew {{include.modifier}}install {{include.item}} 
+{%- endcapture -%} 
 
 Install command:
-<div class="inline">
-    {% highlight bash %}
-    $ {{- code -}}
-    {% endhighlight %}
-
-    <button
-        id="copyButton"
-        aria-label="Copy to clipboard"
-        style="visibility: hidden"
-        onclick="navigator.clipboard.writeText('{{code}}')"
-    >
-    📋
-    </button>
+<div class="copyable">
+    {% highlight bash %} $ {{ code }} {% endhighlight %}
 </div>
 
-<script>
-    if (navigator.clipboard) {
-        document
-            .getElementById("copyButton")
-            .style.removeProperty("visibility");
-    }
-</script>
+<script src="//brew.sh/assets/js/copy.js"></script>
--- a/_includes/install_command.html
+++ b/_includes/install_command.html
@@ -1,32 +1,44 @@
 {%- capture code -%}
 brew {{include.modifier}} install {{include.item}}
 {%- endcapture -%}
-
 <style>
     .inline {
         display: inline-flex;
+        background: rgba(0, 0, 0, 0.3);
+        border-radius: 0.4em;
+        padding-right: 8px;
     }
 
+    .inline .highlight {
+        margin: 0;
+    }
+    
+    .inline .highlight pre {
+        margin: 0;
+    }
+    
     #copyButton {
         background: transparent;
         border: none;
+        border-left: 1px solid #aaa;
     }
 </style>
 
 Install command:
 <div class="inline">
     {% highlight bash %}
-    $ {{code}}
+    $ {{- code -}}
     {% endhighlight %}
-</div>
-<button
-    id="copyButton"
-    aria-label="Copy to clipboard"
-    style="visibility: hidden"
-    onclick="navigator.clipboard.writeText('{{code}}')"
->
+
+    <button
+        id="copyButton"
+        aria-label="Copy to clipboard"
+        style="visibility: hidden"
+        onclick="navigator.clipboard.writeText('{{code}}')"
+    >
     📋
-</button>
+    </button>
+</div>
 
 <script>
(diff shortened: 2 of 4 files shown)


_layouts/cask.html
```diff
@@ -6,6 +6,13 @@
 {%- assign formula_path = "formula" -%}
 {%- assign c = site.data.cask[token] -%}
 <h2>{{ token }}</h2>
+
+{%- capture code -%}
+brew install cask {{ token }}
```
reitermarkus, line 11:
```suggestion
brew install {{ token }}
```

sidharthv96: Should be `brew cask install {{ token }}` right?

image

_includes/code_snippet.html
```diff
@@ -0,0 +1,17 @@
+{% assign code = include.code %}
+<style>
+.copyButton {
+	background: transparent;
+    border: none;
+}
+</style>
+<script src="https://cdn.jsdelivr.net/npm/clipboard@2.0.6/dist/clipboard.min.js"></script>
```
MikeMcQuaid, line 8: This seems like a lot of code and pinned to a version that may become outdated. Could we use our own code here instead?

MikeMcQuaid (commented):
Nice work so far!

_includes/code_snippet.html
sidharthv96: Sure, I think `navigator.clipboard.writeText("{{code}}");` might just work.
 
It's working in a plain html page.

`bundle exec jekyll serve --incremental` takes really long to serve changes, any tips to speed that up?

sidharthv96: That was really an unnecessary library for this usecase.

A single line of custom code worked perfectly in Chrome, FF and Safari!

Nice catch! 😄 

```js
document.querySelector("#copyButton")
    .addEventListener("click", () => navigator.clipboard.writeText("{{code}}"));
```

MikeMcQuaid: > `bundle exec jekyll serve --incremental` takes really long to serve changes, any tips to speed that up?

Could delete various files here.




> Sure, I think `navigator.clipboard.writeText("{{code}}");` might just work.

Sounds good 👍🏻

MikeMcQuaid: Looks mostly good: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard

Should be able to stop it breaking on IE by only showing the icon if `navigator.clipboard` is truthy.

sidharthv96: Are we supporting IE? 

I assumed the targeted users of brew won't be using IE

sidharthv96: Done.

```js
    if(!navigator.clipboard) {
        document.getElementById("copyButton").remove();
    }
```

_layouts/cask.html
reitermarkus: You're right. I got confused because `upgrade` and `reinstall` now also work for casks, `install` doesn't yet.

_includes/code_snippet.html
```diff
@@ -0,0 +1,23 @@
...
+    id="copyButton"
+    aria-label="Copy to clipboard"
+    onclick="navigator.clipboard.writeText('{{code}}')"
+>
+    📋
+</button>
+
+<script>
+    if(!navigator.clipboard) {
+        document.getElementById("copyButton").remove();
```
MikeMcQuaid, line 20: This won't work if JavaScript is disabled. How having a hidden element with a given `class` which JS is run to make unhidden and add the onclick event to?

MikeMcQuaid (commented):
Nice work so far!

_layouts/cask.html
```diff
@@ -6,6 +6,13 @@
 {%- assign formula_path = "formula" -%}
 {%- assign c = site.data.cask[token] -%}
 <h2>{{ token }}</h2>
+
+{%- capture code -%}
+brew cask install {{ token }}
```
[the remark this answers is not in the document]
sidharthv96: https://jekyllrb.com/docs/includes/#passing-parameter-variables-to-includes

Says that we can't have brackets in the params.

Can you share an example?

_includes/code_snippet.html
sidharthv96: Done!

_layouts/cask.html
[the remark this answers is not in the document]
MikeMcQuaid: Could pass `token` as another parameter.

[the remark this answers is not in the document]
sidharthv96: Then we would have to pass if it's cask or not too, then build the command inside there.

The code_snippet was meant to be a generic file and not know about anything else.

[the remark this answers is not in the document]
MikeMcQuaid: I think it's worth decreasing the genericness here until it's needed (which will help avoid a bit of repetition I think too).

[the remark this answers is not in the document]
sidharthv96: Done.

_includes/install_command.html
```diff
@@ -0,0 +1,28 @@
...
+    onclick="navigator.clipboard.writeText('{{code}}')"
+>
+    📋
+</button>
+
+<script>
+    if (navigator.clipboard) {
+        document
+            .getElementById("copyButton")
+            .style.removeProperty("visibility");
```
MikeMcQuaid, line 26:
```suggestion
        document.getElementById("copyButton")
                .style.removeProperty("visibility");
```

Also, this didn't seem to work for me in my local testing. Perhaps this should be a function called by an `onload`?

_includes/install_command.html
```diff
@@ -0,0 +1,28 @@
+{%- capture code -%}
+brew {{include.modifier}} install {{include.item}}
+{%- endcapture -%}
+
+<style>
+    #copyButton {
```
MikeMcQuaid, line 6: Styling wise this looks alright but might be nice to have a background like the installation command on https://brew.sh and some label that explains what this text is.

MikeMcQuaid (commented):
Looking good! Some more comments.

_includes/install_command.html
sidharthv96: image

Does this look good?

MikeMcQuaid: Yes, that's looking much better! How about putting the 📋 inside the background area?

sidharthv96: image

_includes/install_command.html
sidharthv96: Yeah, that was another issue. It only removes the style defined on the element and not in CSS.
This has been fixed in the latest update.

_includes/install_command.html
sidharthv96: image
image

I had tried that first, but it's displayed weird. 

My CSS is a bit rusty any pointers?

sidharthv96: image

Which component is being used here?

MikeMcQuaid: > I had tried that first, but it's displayed weird.
> 
> My CSS is a bit rusty any pointers?

I could try and help once it's committed here. Not sure off the top of my head.



> Which component is being used here?

All the HTML/CSS for that is in https://github.com/homebrew/brew.sh. Hopefully that helps. It would be nice to match that, I agree. We might want to move more CSS into https://github.com/homebrew/brew.sh (as it's shared on this site)

sidharthv96: I've used the syntax highlights same as brew.sh now. 

Can you share which files I should modify to speed up the start time?

It's taking upwards of 30minutes here for every build 😞 .

sidharthv96: image

This looks nice too, maybe we can add a border similar to that around the button?

image

But that's all adding more code, or should we just keep it simple?

sidharthv96: image

Adding a line before the icon should make it okay right?

sidharthv96: image

MikeMcQuaid: > I've used the syntax highlights same as brew.sh now.

Nice!




> Can you share which files I should modify to speed up the start time?

If you rebase on `origin/master` you should find `rake build` and `rake serve` should run a bit quicker and without depending on each other.



> Adding a line before the icon should make it okay right?

I think it's probably nicer without the border?

_includes/install_command.html
```diff
@@ -0,0 +1,49 @@
+{%- capture code -%}
+brew {{include.modifier}} install {{include.item}}
+{%- endcapture -%}
+<style>
+    .inline {
+        display: inline-flex;
+        background: rgba(0, 0, 0, 0.3);
```
MikeMcQuaid, line 7: Would be good to use the existing code here (by adding e.g. a new class to the brew.sh CSS) rather than adding hardcoding the same here.

_includes/install_command.html
```diff
@@ -0,0 +1,49 @@
...
+Install command:
+<div class="inline">
+    {% highlight bash %}
+    $ {{- code -}}
+    {% endhighlight %}
+
+    <button
+        id="copyButton"
+        aria-label="Copy to clipboard"
+        style="visibility: hidden"
```
MikeMcQuaid, line 36: Feels like this should be defined in the CSS?

sidharthv96: If it is in CSS like it was before, then `document.getElementById("copyButton").style.removeProperty("visibility");` won't work, as it is modifying the style on the actual element.

_includes/install_command.html
sidharthv96: A separation is good for user context too right? Otherwise they might just copy the entire line including the paste unicode and paste it in the shell.

_includes/install_command.html
MikeMcQuaid: What about using `display: none` instead?

SMillerDev:
While installing some machines yesterday I was wondering if, after this is done, you'd be willing to do the same for the install command on brew.sh.

MikeMcQuaid:
> While installing some machines yesterday I was wondering if, after this is done, you'd be willing to do the same for the install command on brew.sh.

Yes, great idea 👍🏻. It may even be easier to do that first, actually, because then the JS and CSS could be shared on both?

sidharthv96:
> While installing some machines yesterday I was wondering if, after this is done, you'd be willing to do the same for the install command on brew.sh.

Do you mean this one?

image


Or these? 

image


@MikeMcQuaid In both cases we would have to make the code copy part generic. Unless we plan to allow copy on only the `$ brew install wget` and `$ brew cask install firefox`.

But just those having copy might stand out among the rest.

SMillerDev:
Just the first one.

MikeMcQuaid:
Yup, the first one 👍🏻.

I'm thinking that it's generic based on a element class which then sets the correct CSS and handles it with JS accordingly.

sidharthv96:
Then we'd have to make the install_command.html back to a generic one right?

MikeMcQuaid:
Install command can stay pretty much as-is but the JS/CSS would move to brew.sh.

sidharthv96:
So just the styling and not the copy functionality?

I thought @SMillerDev meant the copy button too.

sidharthv96:
image

Wouldn't this be better ?

MikeMcQuaid:
Yes, sorry, we all agree: that'd be good (except the white stripe)

sidharthv96:
So to sum up,

- All the CSS will be moved to brew.sh/assets/css/styles.scss
- There will be two install_command.html, in each repo.
- The one in brew.sh will just take any string and display it. (There will NOT be copy support in any other install commands listed underneath, just the main install command at the top)
- The one in formulae.brew.sh will be parametrised as discussed here.

sidharthv96:
image
image

I personally think the separator looks good.

MikeMcQuaid:
> * All the CSS will be moved to brew.sh/assets/css/styles.scss

Yup 👍🏻 



> * There will be two install_command.html, in each repo.

I think it shouldn't be needed in brew.sh as it'll only be used in a single place with no `install:` prefix or anything.

> * The one in brew.sh will just take any string and display it. (There will NOT be copy support in any other install commands listed underneath, just the main install command at the top)

Instead: just have a class (e.g. `copyable` or something) which adds the relevant icon and copy-paste logic using JS. Adding the icon with JS instead of removing/unhiding it should make the backwards compatibility easier 👍🏻 





> * The one in formulae.brew.sh will be parametrised as discussed here.

Yup!



> I personally think the separator looks good.

I'm game for a separator if it's a very different colour (e.g. one of the others on the page: yellow or light brown).

_includes/install_command.html
```diff
@@ -0,0 +1,10 @@
+{%- capture code -%} 
+brew {{include.modifier}}install {{include.item}} 
+{%- endcapture -%} 
+
+Install command:
+<div class="copyable">
+    {% highlight bash %} $ {{ code }} {% endhighlight %}
+</div>
+
+<script src="//brew.sh/assets/js/copy.js"></script>
```
MikeMcQuaid, line 10: Needs a newline at the bottom here. Also, I think having this script tag in the base layout on brew.sh would mean it's not needed here.

MikeMcQuaid (commented):
Looking good, almost there!

_includes/install_command.html
sidharthv96: `<script src="{{ "/assets/js/copy.js" | relative_url }}" async></script>`

This should be added to the base.html in brew.sh?

MikeMcQuaid:
Great work here too @sidharthv96, you rock SO HARD 🎉

PR #337, opened 2020-10-01
Simplify loading casks.

reitermarkus:
Use `Cask::CaskLoader::load` instead of creating a `Cask::CaskLoader::FromPathLoader` instance manually.


MikeMcQuaid:
Thanks @reitermarkus ❤️

PR #349, opened 2020-11-08
Make dollar sign inside `.copyable` segments unselectable

vladimyr:
This depends on https://github.com/Homebrew/brew.sh/pull/536


MikeMcQuaid:
Thanks @vladimyr!

PR #420, opened 2021-02-12
maintenance.json: update from 2.7.1 to 2.7.5

nlhomme:
Follows [pull request #99592](https://github.com/Homebrew/homebrew-cask/pull/99592)


Bo98:
This is done automatically by bot every hour. Any changes done outside that would just be reverted by the bot an hour later.

The problem is that the bot currently runs on Catalina, which is why this update was not detected. [Capacity for Big Sur is low](https://github.com/actions/virtual-environments/issues/2486) but I'll look and see if there's anything we can do.

Unfortunately, manually editing it won't work.

PR #450, opened 2021-04-12
Add days to analytics page titles

Rylan12:
Change analytics page titles from "Formula Install Events" to "Formula Install Events (30 days)". This is to support https://github.com/algolia/docsearch-configs/pull/3945. That way, the days are identifiable from the title without needing to read farther (and look at the `h3` `/api` url)

Before:

Screen Shot 2021-04-12 at 12 48 42 PM

After:

Screen Shot 2021-04-12 at 12 48 36 PM

### DIFF
--- a/_layouts/analytics.html
+++ b/_layouts/analytics.html
@@ -8,7 +8,7 @@ layout: base
 {%- else -%}
     {%- assign formula_path = "formula" -%}
 {%- endif -%}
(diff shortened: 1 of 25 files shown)


SMillerDev (approved):

EricFromCanada (approved):

MikeMcQuaid:
Makes sense, thanks @Rylan12!

PR #486, opened 2021-06-08
docs: add bottle API docs

Rylan12:
Follow-up to #482

Adds the new bottle API to the [JSON API Documentation](https://formulae.brew.sh/docs/api/) page.


SMillerDev (approved):

MikeMcQuaid (approved):
This looks good but let's hold off documenting these until we're a bit further along with the feature and are sure we've got the API fully nailed down.

docs/api.md
```diff
@@ -285,6 +285,108 @@ GET https://formulae.brew.sh/api/formula-linux/${FORMULA}.json
...
+- `${FORMULA}`: the name of the formula, e.g. `wget`
+
+#### [Response](https://formulae.brew.sh/api/bottle/wget.json)
+```json
+{
+  "name": "wget",
+  "bottles": {
+    "arm64_big_sur": {
+      "url": "https://ghcr.io/v2/homebrew/core/wget/blobs/sha256:e9034fc9062d5d28972135be031876672aff18fa945ce37e9c2ee1e2c4287f3a"
+    },
```
MikeMcQuaid, line 306:
```suggestion
    "arm64_big_sur": "https://ghcr.io/v2/homebrew/core/wget/blobs/sha256:e9034fc9062d5d28972135be031876672aff18fa945ce37e9c2ee1e2c4287f3a",
```

I almost wonder if this syntax would be a nicer shorthand for when the `sha256` isn't specified? There's going to need to be a different parse path for handling the missing `sha256` case anyway?

MikeMcQuaid: Or even:

```suggestion
    "arm64_big_sur": "$BLOBS/e9034fc9062d5d28972135be031876672aff18fa945ce37e9c2ee1e2c4287f3a",
```

given that the string prefix is identical every time.

Rylan12:
I'm cool with both of those suggestions, I think, but good point about waiting. Also, who knows whether this will need to change in other ways as well.

MikeMcQuaid (approved):

Rylan12:
I updated this because there have been some additions/changes to the API that I don't want to forget about, but I'm still going to hold off on merging until `HOMEBREW_INSTALL_FROM_API` is documented. That way we can make quick iterations as needed (like we've done with `/api/versions-casks.json` and I'm anticipating may happening with `/api/bottle/foo.json`

MikeMcQuaid:
> to hold off on merging until `HOMEBREW_INSTALL_FROM_API` is documented

I think it may be worth holding off merging until not only it's documented but perhaps it's public entirely?

Rylan12:
I was thinking that documenting made it public (i.e. officially supported so we can't just tell people "this isn't supported yet" or something like that), just not necessarily the default. Either way, I'm fine with waiting. I don't think there's any rush to get this merged.

Matausi29 (approved):

PR #496, opened 2021-07-01
formulae: add a disable time and reason to the title

SMillerDev:
Wasn't able to test this locally because it does not want to build.


MikeMcQuaid (approved):
Perfect, thanks @SMillerDev!

PR #521, opened 2021-07-30
docs: stabilize bottle API responses

skyzh:
This PR updates sample API response in the API docs. It might include some changes in https://github.com/Homebrew/formulae.brew.sh/pull/486. Maybe we should get that merged before this PR gets merged.

Regarding the previous discussions about making a Homebrew mirror https://github.com/Homebrew/discussions/discussions/1906, this is our first step to make mirroring easier by stabilizing some fields in JSON API.

Mirror tools made by mirror maintainers from Linux user groups in China require specific fields from the JSON API for the mirroring tools to work. Specifically, they need the following structures:

* JSON API response should at least contains the following structure (or fields):
    ```
    Response = [Formula*]
    Formula = { name, Versions, Bottle, revision }
    Bottle = BottleStable?
    BottleStable = { rebuild, Files }
    Versions = { stable?, bottle }
    Files = { [system_arch]: { url, sha256 } }
    ```
* The `url` field should always contain a blob file on `ghcr.io`, like `https://ghcr.io/v2/homebrew/core/wget/blobs/sha256:ab03f58f6d9a4018b1b0bfd53e5b797fcd90b86b1d60b20948de499ec4d4c6b4`.

Reference: [USTCLUG's bottles-json parser](https://github.com/ustclug/ustcmirror-images/blob/master/homebrew-bottles/bottles-json/src/main.rs), [SJTUG's mirror-clone](https://github.com/sjtug/mirror-clone/blob/master/src/homebrew.rs)

This PR just serves as a hint on how mirror tools use Homebrew APIs. If any contributor needs to update the response structure of Homebrew APIs, this PR would make them aware that changing some specific field will break mirror tools in the community unintentionally.

cc @taoky @shankerwangmiao does this look okay to you?

And cc @MikeMcQuaid for review, many thanks!

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -48,26 +48,33 @@ GET https://formulae.brew.sh/api/cask.json
     "version_scheme": 0,
     "bottle": {
       "stable": {
-        "rebuild": 0,
-        "cellar": "/usr/local/Cellar",
-        "prefix": "/usr/local",
-        "root_url": "https://homebrew.bintray.com/bottles",
+        "rebuild": 1,
+        "root_url": "https://ghcr.io/v2/homebrew/core",
         "files": {
           "arm64_big_sur": {
-            "url": "https://homebrew.bintray.com/bottles/wget-1.21.1.arm64_big_sur.bottle.tar.gz",
-            "sha256": "e9034fc9062d5d28972135be031876672aff18fa945ce37e9c2ee1e2c4287f3a"
+            "cellar": "/opt/homebrew/Cellar",
+            "url": "https://ghcr.io/v2/homebrew/core/wget/blobs/sha256:ab03f58f6d9a4018b1b0bfd53e5b797fcd90b86b1d60b20948de499ec4d4c6b4",
+            "sha256": "ab03f58f6d9a4018b1b0bfd53e5b797fcd90b86b1d60b20948de499ec4d4c6b4"
           },
           "big_sur": {
-            "url": "https://homebrew.bintray.com/bottles/wget-1.21.1.big_sur.bottle.tar.gz",
-            "sha256": "e6ea2a50b8196206f7072360e713535bb16fb786c8b5fe23cab05757e0f67b13"
+            "cellar": "/usr/local/Cellar",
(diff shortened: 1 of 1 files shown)


Rylan12 (approved):
Good call, thanks @skyzh!

Thinking about keeping this updated in the future: because changes to this output are made in Homebrew/brew, it's much less likely for contributors to remember to update the docs here. You could, if you wanted, use `brew info --json` (and its various invocations) as the main point of reference moving forward because that's more accurate to this output in terms of the fields present. It's not perfect, though, because the top-level format (i.e. is it an array or an object) doesn't quite match here, and you need to use `--json=v2` for casks.

I wonder if we can automatically update this every time it changes. I'll look into it and maybe do that in #486 (which is intentionally blocked at the moment). I'm thinking something like how we can access `{{ site.taps.core.repo }}` on this page. Not a blocker, here, just mentioning it in case anyone has thoughts.

Rylan12:
Thanks, @skyzh!

Just so you know, I've opened #523 which, if merged, will automatically update these samples which will keep them up to date (and mean we won't need PRs like this every few months when something changes).

Rylan12:
Just an update: #523 has been merged so hopefully, this won't be a problem in the future. 🤞 Plus, in addition to updating documentation automatically, we're probably going to add additional checks to [make sure that we don't make backward-incompatible changes accidentally in the future](https://github.com/Homebrew/formulae.brew.sh/pull/523#issuecomment-892064105). Not sure the timeline for that, yet, but keep your eyes peeled if you're curious.

skyzh:
> Just an update: #523 has been merged so hopefully, this won't be a problem in the future. 🤞 Plus, in addition to updating documentation automatically, we're probably going to add additional checks to [make sure that we don't make backward-incompatible changes accidentally in the future](https://github.com/Homebrew/formulae.brew.sh/pull/523#issuecomment-892064105). Not sure the timeline for that, yet, but keep your eyes peeled if you're curious.

Thanks a lot! I'll keep an eye on that.

Issue #566, opened 2021-10-21
Migrate to homebrew-core for Linux


MikeMcQuaid:
We've now migrated Linux formulae from linuxbrew-core to homebrew-core. Unfortunately, this breaks a non-trivial number of assumptions in this codebase.

Things that need to be updated (in a rough order, combining/jumping steps is fine):

- [x] 1. the remote used for generating/describing/linking Linux formulae needs to be homebrew-core rather than linuxbrew-core
- [x] 2. the Linux formulae should share a formula page with macOS formulae and the Linux bottle used to detect Linux support
- [x] 3. the Linux/macOS dependencies (and perhaps: caveats, anything else?) should be separated when they differ between OSs (but still displayed on the same page)
  - [x] this requires updating the formula API to differentiate between macOS and Linux dependencies, conflicting formulae, `keg_only` / deprecated / disabled status, and (possibly) caveat text
- [x] 4. the Linux analytics URLs should reference "homebrew-core" not "linuxbrew-core"
- [x] 5. use the Jekyll `redirect_from` plugin to ensure that we don't break any links that used to work (e.g. migrate API JSON, formula/analytics/index pages)

Things that don't need updating:
- analytics data: macOS and Linux are still (for now) using separate analytics data

CC @EricFromCanada @Rylan12 @issyl0 for folks who have done Linux work here previously and may be interested in volunteering ❤️

issyl0:
I am interested in doing at least some of this, so I'll try to carve out time. With packing and moving house in the next few weeks I'm not sure how much time is left.

EricFromCanada:
Since each formula's [JSON file](https://formulae.brew.sh/api/formula/coreutils.json) will need to also contain the Linux statistics from its correpsonding [formula-linux JSON file](https://formulae.brew.sh/api/formula-linux/coreutils.json), I'm going to assume tacking them on as an `analytics-linux` key is the best approach.

MikeMcQuaid:
@EricFromCanada makes sense for me!

SMillerDev:
Isn't nr. 3 done by @Rylan12 for the Homebrew API?

MikeMcQuaid:
@SMillerDev unsure!

Bo98:
I think the HTML pages still need updating at least.

EricFromCanada:
☝️ Updating the Liquid templates is on my list

Rylan12:
[...]

PR #569, opened 2021-10-23
Migrate Linux data source to Homebrew/homebrew-core

issyl0:
- Very WIP, opening this as very draft to a) chart my progress and b) give visibility to others who might want to do parts of #566.
- At the moment this is just me messing around trying to remember what I did here when I implemented Linux formulae on this site - that was two years ago! :scream:

My vague plan here:

- The low-hanging fruit, README, index page.
- Remove `formula-linux/` and see what breaks. (This is where I currently am at the time of writing this.)
- Check we've got all the necessary Linux things on the macOS (I should say "formula") pages now.
- The rest of the things in #566, unless anyone else gets to them first.

Jekyll is still a mystery to me, so I can't promise this'll be fast.


issyl0:
Thanks for the help, Eric, and for ticking the checklist items in the issue. There's quite a lot complete here! ✨

Over this weekend I'll work on item 2, the formula page: splitting up the views by macOS and Linux dependencies/caveats/bottles and the analytics tables.

issyl0:
On the "show macOS and Linux analytics in separate tables", I'd quite like to show them in one table with multiple columns inside for macOS and Linux, but I couldn't work out how to do it with all the loops and different paths for `analytics` and `analytics-linux`. Maybe that can be a future goal!

EricFromCanada:
The issue there is how the stats table markup iterates through each formula's invocation form that appears in each data source, whether just the formula name or with options. To combine the tables would require creating a merged list of these from each data source before fetching the values for each form, if they exist. But, since in core the only possibilities are either `[formula]` or `[formula] --HEAD`, the loop could check for each form on every loop and only print the row if a value was present in either data source.

EricFromCanada:
With for loops implemented and `analytics-linux` added, that's a wrap for the JSON templates. Although the same kind of looping could be added to the formula HTML template, it probably needs a complete redesign.

issyl0:
> 3. the Linux/macOS dependencies, analytics (and perhaps: caveats, anything else?) should be separated when they differ between OSs (but still displayed on the same page). This requires updating the formula API to differentiate between macOS and Linux dependencies, conflicting formulae, keg_only status, and (possibly) caveat text.

This is the last thing to do here, and it feels quite big, and this PR is extremely large already. I know if we don't ship step 3 with steps 1, 2, 4 and 5 we'll be lacking functionality for a bit. But if people are pulling Linux data from this API now they're getting out of date stuff anyway). All that to say, does anyone else have more time than me to work on this part? @Rylan12?

SMillerDev:
Am I missing something or is the easiest solution to stop splitting Linux and macOS analytics reporting?

EricFromCanada:
> Am I missing something or is the easiest solution to stop splitting Linux and macOS analytics reporting?

No, that part is done. The issue now is that anything in formulae that's potentially affected by an `on_linux` block isn't reflected in its JSON representation, so the formula API (and therefore the formulae.brew.sh pages) are showing Mac-only info. I'd say we're good to ship what we have and follow up with more once the formula API is updated

Rylan12:
> This is the last thing to do here, and it feels quite big, and this PR is extremely large already. I know if we don't ship step 3 with steps 1, 2, 4 and 5 we'll be lacking functionality for a bit. But if people are pulling Linux data from this API now they're getting out of date stuff anyway). All that to say, does anyone else have more time than me to work on this part? @Rylan12?

Eventually, yes, I'd be happy to work on this a bit. At the moment, though, I'm a bit swamped with `HOMEBREW_INSTALL_FROM_API` issues and school (midterms start again soon 😭). If I end up with some more time and start to work on it, I'll leave a comment.

MikeMcQuaid:
> This is the last thing to do here, and it feels quite big, and this PR is extremely large already. I know if we don't ship step 3 with steps 1, 2, 4 and 5 we'll be lacking functionality for a bit. But if people are pulling Linux data from this API now they're getting out of date stuff anyway). All that to say, does anyone else have more time than me to work on this part? @Rylan12?

Yes, I'm fine to ship without 3 for sure.

MikeMcQuaid (approved):
❤️. Rebase this and merge whenever you get a chance (because there will be inevitable conflicts soon after otherwise).

issyl0:
I think GitHub is confused by the sheer size of this PR. There were no conflicts when I rebased on `master` locally.

SMillerDev:
Thanks @issyl0 and @EricFromCanada  ! Without your contributions it'd be impossible to keep homebrew going. You can feel good knowing that you've made the world a lot better for homebrew users around the world! 👍 🎉

Rylan12:
> I think GitHub is confused by the sheer size of this PR. There were no conflicts when I rebased on `master` locally.

What I usually do to avoid problems like this is only commit the changes that are to the generation files and not the actual site files. Then, once the PR is merged, the automatic generation is triggered and the updated site is pushed as an additional commit.

PR #654, opened 2022-06-17
Add `/api/cask-source.json` endpoint

Rylan12:
This PR adds a new endpoint that contains the source code for every cask. This will be used when loading casks from the API.


Bo98:
This does start to feel like something that would be better more like the formula way of a proper JSON representation and converting it to code within `brew`, but the Cask DSL probably isn't quite there with the conditionals etc it has.

Rylan12:
Passing on this for now

PR #658, opened 2022-07-23
Use `#to_hash_with_variations` to generate formula and cask JSON

Rylan12:
Now that we can create `variations` hashes for formulae and casks, let's include them here in the JSON API. For a little bit, the cask API did include the variations hash, but that was changed when we started only doing the `variations` calculations when `--variations` was passed to `brew info --json`. This PR re-adds this functionality and adds it for formulae.


SMillerDev (approved):

MikeMcQuaid:
Woohoo, nice work @Rylan12!

FYI @EricFromCanada that now we have these variations it'd be nice to display them in the UI!

PR #660, opened 2022-09-07
Update visit.json

pnavaro:
There is a sha256 mismatch here. I replaced it by the sequence you can find in https://github.com/visit-dav/visit/releases/download/v3.3.0/visit_sha256_checksums.txt


MikeMcQuaid:
You need to make this change in Homebrew/homebrew-core.

PR #664, opened 2022-10-07
scheduled: use macos-11.

MikeMcQuaid:
Let's see if this fixes the regeneration issues.


MikeMcQuaid:
This doesn't fix things.

PR #665, opened 2022-10-07
scheduled: use macOS 12.

MikeMcQuaid:
Attempt #664 again but with macOS 12 instead of 11.


p-linnane:
Is it worth reverting https://github.com/Homebrew/homebrew-cask/commit/83672e7493b2ce404bd128d5f73f5e5e861adcc8? Seems to be another side effect of the `on_{macos_version}` conversion that was started.

PR #673, opened 2022-12-06
Use `-` instead of `HEAD` in GitHub URLs

nihaals:
`HEAD` gives a slightly weird experience on GitHub as it:
- Shows the current commit hash in some places (like the branch picker)
- Doesn't show the current branch
- Will just switch to using `HEAD`'s commit hash when using relative links (like click the name of the parent path)
  - This means future links are connected to a specific commit unnecessarily instead of the branch

Using `-` instead avoids hardcoding the name of the default branch while making the behaviour nicer when actually on GitHub as it simply redirects to the default branch. This keeps the same behaviour as `HEAD` but avoids the disadvantages mentioned above.


SMillerDev (approved):
Thanks @nihaals  ! Without contributions like yours it'd be impossible to keep homebrew going with the high standards that users have come to expect from the project. You can feel good knowing that you've made the world a tiny bit better for homebrew users around the world! 👍 🎉

PR #678, opened 2023-01-12
Remove scope attribute from td elements

samford:
Removing the `scope` attribute from `td` elements addresses the following HTML validation error:

> The "scope" attribute on the "td" element is obsolete. Use the "scope" attribute on a "th" element instead.

The parent `th` elements already use a `scope` attribute and this change appears to render the same when tested in current browsers.


Rylan12 (approved):

MikeMcQuaid:
Seems fine, thanks.

PR #679, opened 2023-01-12
Escape formula name in templates

samford:
A few items in the analytics contain the text for an `Options` object, like `adns #<Options:0x0000000102592c90>`. When used in related HTML templates, we end up with `<code>adns #<Options:0x0000000102592c90></code>`, which produces various validation errors because the `Options` text is being treated as an element:

> Element "options:0x0000000102592c90" not allowed as child of element "code" in this context.

> End tag "code" violates nesting rules.

This PR resolves the errors by escaping the formula name in templates, so the related HTML is `<code>adns #&lt;Options:0x0000000102592c90&gt;</code>` instead.

There may be a better way to address this when generating the related JSON data (e.g., `_data/analytics/build-error/365d.json`) but escaping this in the HTML templates may be a good idea regardless.

### DIFF
--- a/_layouts/analytics.html
+++ b/_layouts/analytics.html
@@ -40,9 +40,9 @@ layout: base
         {%- assign data_name = full_name | remove: "@" | remove: "." | replace: "+", "_" -%}
         {%- assign formula = site.data.formula[data_name] -%}
         {%- if full_name == formula.name -%}
-            <a href="{{ site.baseurl }}/formula/{{ full_name }}"><code>{{ item.formula }}</code></a>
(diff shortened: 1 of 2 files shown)


SMillerDev (approved):

Rylan12 (approved):

MikeMcQuaid (approved):

MikeMcQuaid:
> There may be a better way to address this when generating the related JSON data (e.g., `_data/analytics/build-error/365d.json`)

Yes please, would be good to pick this up, thanks @samford!

PR #680, opened 2023-01-12
Naively remove ANSI color codes from caveats

samford:
The `vapoursynth` formula currently uses ANSI color codes in the `caveats` text to make certain text bold (e.g., `\x1B[3m\x1B[1mvapoursynth.core.sub\x1B[0m`) and this leads to a `Forbidden code point U+001b` HTML validation error.

This PR addresses the issue by naively removing ANSI color codes from formula caveats text when generating formula data. This approach definitely isn't exhaustive (I'm open to better alternatives) but it works for the codes currently used in `vapoursynth`, at least.

Alternatively, we can simply remove the ANSI codes from the `vapoursynth` formula's `caveats` text and discourage further use. This is currently the only formula/cask doing this, so it doesn't appear to be a widespread issue.

### DIFF
--- a/script/generate.rb
+++ b/script/generate.rb
@@ -1,4 +1,12 @@
 #!/usr/bin/env brew ruby
+
+def sanitize_formula_text(data)
+  # Naively remove ANSI color codes, to avoid HTML validation errors
+  data["caveats"] = data["caveats"].gsub(/\e\[[0-9;]+m/i, "") if data["caveats"].present?
+
+  data
+end
+
 os = ARGV.first
 tap_name = ARGV.second
 tap = Tap.fetch(tap_name)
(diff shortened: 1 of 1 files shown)


Rylan12 (commented):
I don't think we should sanitize the data here because the JSON API files will contain the sanitized input. I think it would be fine to sanitize with this method post-JSON if that's possible, but I don't think we should be modifying the raw JSON API data for HTML presentation purposes now that we're using the API for installs.

Alternatively, could we selectively ignore the error for formula pages? Or even better, only for the caveat section? That may not be possible, not sure

SMillerDev:
I think we should just remove it and disallow it with rubocop.

MikeMcQuaid:
> I don't think we should sanitize the data here because the JSON API files will contain the sanitized input. I think it would be fine to sanitize with this method post-JSON if that's possible, but I don't think we should be modifying the raw JSON API data for HTML presentation purposes now that we're using the API for installs.

Agreed.



> I think we should just remove it and disallow it with rubocop.

Agreed.

samford:
> I don't think we should sanitize the data here because the JSON API files will contain the sanitized input. I think it would be fine to sanitize with this method post-JSON if that's possible, but I don't think we should be modifying the raw JSON API data for HTML presentation purposes now that we're using the API for installs.

That makes sense. I wasn't sure if keeping the ANSI color codes in the JSON data was preferable or not, so thanks for clarifying. If we went in this direction, it would probably make more sense as a custom Liquid filter (i.e., sanitizing the data in HTML templates only).

That said, I agree with simply removing the ANSI codes from the `vapoursynth` caveats and adding a related Rubocop. I'll open a PR in homebrew-core and look into the Rubocop either later today or this weekend when I have a moment.

MikeMcQuaid:
Thanks @samford!

PR #697, opened 2023-01-30
Allow manual runs of the `scheduled.yml` workflow

Rylan12:
We should be able to manually trigger a re-deploy of the site. For example, if we're making an API change in Homebrew/brew, we may want to have two-step system where we merge an API changed, wait for it to be deployed to the site, and then update the formula/cask loader to handle it. Instead of needing to wait an hour for the update to happen, we should be able to trigger one immediately and not have to wait.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -6,6 +6,7 @@ on:
   schedule:
     # Once every hour
     - cron: "0 * * * *"
+  workflow_dispatch:
 
 concurrency:
   group: scheduled


Bo98 (approved):

SMillerDev (approved):

MikeMcQuaid:
Nice addition!

Ideally we'd automatically trigger something like this when we change homebrew/core or homebrew/cask if it's possible?

Rylan12:
In theory. If we can trigger workflows cross-repo, maybe we should have this be the last step of the scheduled merge in homebrew/core and homebrew/cask. Not sure we need it to run for every single merged PR though. If we do that, we probably don't want this running on a schedule here at all

PR #702, opened 2023-02-02
Remove cask-source, bottle and versions APIs

MikeMcQuaid:
These are no longer needed after https://github.com/Homebrew/brew/pull/14439 and https://github.com/Homebrew/brew/pull/14500 are merged and in a stable tag.


Rylan12 (approved):
Thanks! I need to double-check this but I think we can also remove the `versions-casks` and `versions-formulae` APIs

MikeMcQuaid:
> but I think we can also remove the `versions-casks` and `versions-formulae` APIs

@Rylan12 Good call, done here and in https://github.com/Homebrew/brew/pull/14500.

Rylan12 (approved):
Thanks!

Rylan12 (approved):
Thanks!

Rylan12 (approved):

Rylan12 (approved):

PR #720, opened 2023-02-05
script/generate-api-samples: abort on failures.

MikeMcQuaid:
Let's ensure that we blow up if we don't get what we need.

### DIFF
--- a/script/generate-api-samples.rb
+++ b/script/generate-api-samples.rb
@@ -13,6 +13,14 @@ SAMPLES = {
   versions_formulae:                        "versions-formulae.json",
 }.freeze
 
+def failed!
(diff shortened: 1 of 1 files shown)


Rylan12 (approved):

SMillerDev (approved):

MikeMcQuaid:
@Rylan12 https://github.com/Homebrew/formulae.brew.sh/actions/runs/4098274281/jobs/7067268025 reveals the problem here: these two files don't exist on formulae.brew.sh can we can't generate samples this way until we do. If you can find a way to hardcode them or something before that point: please do! Will look with you properly tomorrow.

PR #736, opened 2023-02-11
layouts: define more data in _config.yml

EricFromCanada:
Fetching loop items from the config file data reduces the number of temporary arrays that Liquid needs to create and seems to speed up generation slightly; `241.446 seconds` vs. `230.459 seconds` in my testing.


MikeMcQuaid (approved):

MikeMcQuaid:
Works for me, nice optimisation @EricFromCanada!

PR #744, opened 2023-02-13
generate-cask: enable `Cask.generating_hash!` mode

MikeMcQuaid:
See https://github.com/Homebrew/brew/pull/14615

This will set placeholder values in the generated hash.


Bo98:
I don't understand this side since https://github.com/Homebrew/brew/pull/14615 is calling this itself inside `to_h`.

Bo98:
Ok, that makes sense.

PR #752, opened 2023-02-15
scheduled: update every 15m.

MikeMcQuaid:
The job now takes ~8m so this gives us a bit of room while increasing the frequency of updates 4x.


Bo98:
The Algolia crawl takes longer than 15 minutes so maybe we should recrawl on a slower interval.

PR #758, opened 2023-02-16
Re-add cask-source API documentation

MikeMcQuaid:
If we want to let others use this too.


MikeMcQuaid:
Passing on this in case we kill this API again in future.

PR #760, opened 2023-02-16
workflows/scheduled: retry analytics query.

MikeMcQuaid:
This sometimes fails so let's retry it.


Bo98:
`rake all_analytics` does not attempt a regeneration if the 30d build-error part has already been generated, which happens to be the first file it generates. So if it fails after first JSON then what is actually happens here is we're ignoring it and uploading a partial dataset.

Instead of changing the workflow I recommend focussing a retry on the individual `sh` command invokation. If that fails again then it should continue to abort.

MikeMcQuaid:
@Bo98 fair points. Addressed in https://github.com/Homebrew/formulae.brew.sh/pull/761

PR #761, opened 2023-02-16
Fix analytics retries

MikeMcQuaid:
Move the retry logic to the `Rakefile` instead.

While we're here, cleanup some unneeded analytics checks.


Bo98 (approved):

Bo98:
Thanks!

PR #766, opened 2023-02-17
docs/api: clarify cask info

p-linnane:
Minor readability fixes for clarity.


MikeMcQuaid:
Good catch, thanks @p-linnane!

PR #771, opened 2023-02-20
Add signed API endpoints

Bo98:
This uses JSON Web Signatures with a 4096-bit RSA key using RSA-PSS. The JSON structure is defined by RFC7515 (and RFC7797 for the `b64` part).

The endpoints are /api/formula.jws.json and /api/cask.jws.json.

Other endpoints are not signed at this time.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -163,6 +163,12 @@ jobs:
       - name: Validate build
         run: ./script/validate-build.rb
 
+      - name: Sign API
+        env:
+          JWS_SIGNING_KEY_ID: homebrew-1
+          JWS_SIGNING_KEY: ${{ secrets.JWS_HOMEBREW_1 }}
+        run: ./script/sign-json.rb
+
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v1
   deploy:
--- a/script/sign-json.rb
+++ b/script/sign-json.rb
@@ -1,0 +1,57 @@
+#!/usr/bin/env ruby
+require "base64"
+require "json"
+require "openssl"
+require "pathname"
+
+ROOT = Pathname(__dir__).parent.freeze
+
+# If we need to roll keys in the future, we'd likely expand this to take in multiple keys.
+KEY_ID = ENV.fetch("JWS_SIGNING_KEY_ID").freeze
+PRIVATE_KEY = OpenSSL::PKey::RSA.new(ENV.fetch("JWS_SIGNING_KEY")).freeze
+
+[
+  ROOT/"_site/api/formula.json",
+  ROOT/"_site/api/cask.json",
+].each do |path|
+  data_string = path.read
+
+  # The JSON structure throughout this script is a standard format. References:
+  # - RFC7515 [JSON Web Signature (JWS)]
+  # - RFC7797 [JWS Unencoded Payload Option]
+  header = {
+    # We're using RSA-PSS with SHA-512 - see RFC7518 (§3.5) or https://www.iana.org/assignments/jose/jose.xhtml.
+    "alg": "PS512",
+    # Standard JWS uses base64 URL-encoded payloads, but we don't need that and it compresses worse.
+    # RFC7797 (§3) defines the `b64` option here we use.
+    "b64": false,
+    # Required by RFC7797 (§6): this tells clients they should not parse this if they don't understand what b64 means
+    "crit": ["b64"],
+  }
+  header_base64 = Base64.urlsafe_encode64(header.to_json) # Base64URL is required - see RFC7515 (§3).
+
+  # Formula defined by RFC7797 (§3).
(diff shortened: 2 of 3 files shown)


.github/workflows/scheduled.yml
```diff
@@ -163,6 +163,12 @@ jobs:
       - name: Validate build
         run: ./script/validate-build.rb
 
+      - name: Sign API
+        env:
+          JWS_SIGNING_KEY_ID: homebrew-1
+          JWS_SIGNING_KEY: ${{ secrets.JWS_HOMEBREW_1 }}
```
MikeMcQuaid, line 169: Is this setup yet or does it need set?

script/sign-json.rb
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env ruby
```
MikeMcQuaid, line 1: Thoughts on perhaps moving this into Homebrew/brew?

script/sign-json.rb
```diff
@@ -0,0 +1,47 @@
...
+[
+  ROOT/"_site/api/formula.json",
+  ROOT/"_site/api/cask.json",
+].each do |path|
+  data_string = path.read
+
+  # References:
+  # - RFC7515 [JSON Web Signature (JWS)]
+  # - RFC7797 [JWS Unencoded Payload Option]
+
```
MikeMcQuaid, line 20:
```suggestion
  # - RFC7797 [JWS Unencoded Payload Option]
```

script/sign-json.rb
```diff
@@ -0,0 +1,47 @@
...
+  header = {
+    "alg": "PS512",
+    "b64": false,
+    "crit": ["b64"],
```
MikeMcQuaid, line 24: Explaining why each of these values is what it is today would be handy. Maybe worth turning `"b64"` into a variable, too, as it seems to be shared here.

script/sign-json.rb
```diff
@@ -0,0 +1,47 @@
...
+  header_base64 = Base64.urlsafe_encode64(header.to_json)
+
+  signing_input = "#{header_base64}.#{data_string}"
+
+  signature = {
+    "protected": header_base64,
+    "header": {
+      "kid": KEY_ID,
+    },
+    "signature": Base64.urlsafe_encode64(PRIVATE_KEY.sign_pss("SHA512", signing_input, salt_length: :digest, mgf1_hash: "SHA512")),
```
MikeMcQuaid, line 35: Maybe split this out into a variable as well.

MikeMcQuaid (approved):
Wow, this is great! Nice work here.

.github/workflows/scheduled.yml
Bo98: I've added one I generated on my machine. I will destroy the copy on my machine but can put it in 1Password before I do that.

script/sign-json.rb
MikeMcQuaid: Removing whitespace. The diff doesn't seem to show it properly.

.github/workflows/scheduled.yml
MikeMcQuaid: Perfect, thanks and: yes please on 1Password. Please use the Ops vault (rather than Maintainers).

script/sign-json.rb
MikeMcQuaid: Yeh, we should move JSON generation to brew ASAP too IMO. Could write up a `help wanted` issue for that?

script/sign-json.rb
Bo98: Added (probably too much) comments.

script/sign-json.rb
```diff
@@ -0,0 +1,57 @@
...
+].each do |path|
+  data_string = path.read
+
+  # The JSON structure throughout this script is a standard format. References:
+  # - RFC7515 [JSON Web Signature (JWS)]
+  # - RFC7797 [JWS Unencoded Payload Option]
+  header = {
+    # We're using RSA-PSS with SHA-512 - see RFC7518 (§3.5) or https://www.iana.org/assignments/jose/jose.xhtml.
+    "alg": "PS512",
+    # Standard JWS uses base64url-encoded payloads, but we don't need that and it compresses worse.
```
MikeMcQuaid, line 25:
```suggestion
    # Standard JWS uses base64 URL-encoded payloads, but we don't need that and it compresses worse.
```

MikeMcQuaid (approved):
The extra comments are great here actually and exactly what I was hoping for, thanks!

script/sign-json.rb
Rylan12: I opened one here: https://github.com/Homebrew/brew/issues/14730. Feel free to modify if I missed anything

Bo98: Thanks!

Rylan12 (approved):

PR #786, opened 2023-02-26
Use `brew generate-analytics-api` and cleanup `Rakefile`

Rylan12:
Use the new `brew generate-analytics-api` command that will be added in https://github.com/Homebrew/homebrew-formula-analytics/pull/374 (this can't be merged until that is merged)

Also, while we're here, let's clean up the `Rakefile` and GitHub actions workflow a bit:
- Rename the `cask` task to `casks` for consistency
- Remove the option to generate analytics for only macOS/Linux
- Remove the option to generate analytics _and_ either core or cask data in the same task (this just didn't seem necessary)
- Add a new `generate` task that generates the entire site (`formulae`, `casks`, `analytics` and `api_samples`)
- Make the `generate` task the default task
- Remove some unused methods
- Use `bundle exec rake` across the board to call `rake` in GitHub actions (before it was a mix of `bundle exec rake`, `rake` and `/usr/bin/rake`)
- Run `brew update-reset` as its own GitHub actions task before generating core and cask data.

The thinking here is that if someone wants to clone this repo and generate the site data, all they have to do is run `rake` and the default is to generate everything. We still have the ability to generate one at a time in GitHub actions, but I think these defaults make more sense.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -30,10 +30,11 @@ jobs:
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@master
 
+      - name: Reset homebrew/cask tap
+        run: brew update-reset $(brew --repository homebrew/cask)
+
       - name: Update data for homebrew/cask
-        run: |
-          brew update-reset $(brew --repository homebrew/cask)
-          /usr/bin/rake cask
+        run: bundle exec rake casks
         env:
           HOMEBREW_DEVELOPER: 1
 
@@ -60,8 +61,11 @@ jobs:
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@master
 
+      - name: Reset homebrew/core tap
+        run: brew update-reset $(brew --repository homebrew/core)
+
       - name: Update data for homebrew/core
-        run: /usr/bin/rake formulae
+        run: bundle exec rake formulae
         env:
           HOMEBREW_DEVELOPER: 1
 
@@ -99,7 +103,7 @@ jobs:
         run: echo "$ANALYTICS_JSON_KEY" > ~/.homebrew_analytics.json
 
       - name: Update analytics data
-        run: rake all_analytics
+        run: bundle exec rake analytics
 
       - name: Archive data
         run: tar czvf data-analytics.tar.gz _data/analytics _data/analytics-linux api/analytics api/analytics-linux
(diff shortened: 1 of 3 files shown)


.github/workflows/scheduled.yml
```diff
@@ -60,8 +61,11 @@ jobs:
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@master
 
+      - name: Reset homebrew/core tap
+        run: brew update-reset $(brew --repository homebrew/core)
```
MikeMcQuaid, line 65: Are you sure this is necessary? It's really slow.

.github/workflows/scheduled.yml
```diff
@@ -30,10 +30,11 @@ jobs:
...
         uses: Homebrew/actions/setup-homebrew@master
 
+      - name: Reset homebrew/cask tap
+        run: brew update-reset $(brew --repository homebrew/cask)
+
       - name: Update data for homebrew/cask
-        run: |
-          brew update-reset $(brew --repository homebrew/cask)
-          /usr/bin/rake cask
+        run: bundle exec rake casks
```
MikeMcQuaid, line 37: Let's use `/usr/bin/rake` unless it's actually needed, it's a bit faster. If we can use it everywhere: all the better.

Of course, if it's not meaningfully faster and there's some other benefit in being consistent here: ignore me!

MikeMcQuaid (commented):
Looks good so far!

MikeMcQuaid:
@Rylan12 Can you recreate this PR not from a fork so the scheduled job can be tested in the PR? Thanks ❤️

.github/workflows/scheduled.yml
Rylan12: Maybe not, do you think it's necessary for the `homebrew/cask` tap?

.github/workflows/scheduled.yml
Rylan12: Happy to use `/usr/bin/rake` everywhere unless it doesn't work. I just wanted the consistency and chose `bundle exec rake` over the others for no real reason. But yeah, if one doesn't work in a certain place I don't mind the inconcsistency

.github/workflows/scheduled.yml
MikeMcQuaid: @Rylan12 Yeh, it's needed there but not here I think because `setup-homebrew` sets up homebrew/core and not homebrew/cask.

.github/workflows/scheduled.yml
MikeMcQuaid: Yeh, whichever works.

PR #791, opened 2023-02-27
Use `brew generate-analytics-api` and cleanup `Rakefile`

Rylan12:
Remake of https://github.com/Homebrew/formulae.brew.sh/pull/786

Use the new `brew generate-analytics-api` command that will be added in https://github.com/Homebrew/homebrew-formula-analytics/pull/374 (this can't be merged until that is merged)

Also, while we're here, let's clean up the `Rakefile` and GitHub actions workflow a bit:
- Rename the `cask` task to `casks` for consistency
- Remove the option to generate analytics for only macOS/Linux
- Remove the option to generate analytics _and_ either core or cask data in the same task (this just didn't seem necessary)
- Add a new `generate` task that generates the entire site (`formulae`, `casks`, `analytics` and `api_samples`)
- Make the `generate` task the default task
- Remove some unused methods
- Use `/usr/bin/rake` across the board to call `rake` in GitHub actions (before it was a mix of `bundle exec rake`, `rake` and `/usr/bin/rake`)
- Move `brew update-reset` to its own GitHub actions task before generating cask data.

The thinking here is that if someone wants to clone this repo and generate the site data, all they have to do is run `rake` and the default is to generate everything. We still have the ability to generate one at a time in GitHub actions, but I think these defaults make more sense.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -165,7 +165,7 @@ jobs:
           mv -v api-samples _includes/api-samples
 
       - name: Build site
-        run: /usr/bin/rake build
+        run: bundle exec jekyll build
 
       - name: Validate build
         run: ./script/validate-build.rb
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -30,10 +30,11 @@ jobs:
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@master
 
+      - name: Reset homebrew/cask tap
+        run: brew update-reset $(brew --repository homebrew/cask)
+
       - name: Update data for homebrew/cask
-        run: |
-          brew update-reset $(brew --repository homebrew/cask)
-          /usr/bin/rake cask
+        run: /usr/bin/rake casks
         env:
           HOMEBREW_DEVELOPER: 1
 
@@ -99,7 +100,7 @@ jobs:
         run: echo "$ANALYTICS_JSON_KEY" > ~/.homebrew_analytics.json
 
       - name: Update analytics data
-        run: rake all_analytics
+        run: /usr/bin/rake analytics
 
       - name: Archive data
         run: tar czvf data-analytics.tar.gz _data/analytics _data/analytics-linux api/analytics api/analytics-linux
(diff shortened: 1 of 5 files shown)


MikeMcQuaid (approved):
👍🏻 to 🚢 once CI is 🟢 on the scheduled job

.github/workflows/scheduled.yml
```diff
@@ -164,7 +165,7 @@ jobs:
           mv -v api-samples _includes/api-samples
 
       - name: Build site
-        run: bundle exec rake build
+        run: /usr/bin/rake build
```
MikeMcQuaid, line 168:
```suggestion
        run: bundle exec rake build
```

looks like this is the only one that needs it.

could even try:

```suggestion
        run: bundle exec jekyll build
```

and further cleanup the `Rakefile`.

Rakefile
```diff
@@ -145,21 +54,8 @@ task :api_samples do
...
-
-desc "Dump all analytics (macOS and Linux)"
-task all_analytics: :analytics do
-  Rake::Task["analytics"].tap(&:reenable).invoke("linux")
-end
+desc "Generate the API files"
+task generate: %i[formulae casks analytics api_samples]
 
 desc "Build the site"
 task :build do
```
MikeMcQuaid, line 61: Not sure if we're using much of the stuff below? Feels like it could either get run automatically or removed.

Rylan12: Yeah, it's just a question of whether having them in the `Rakefile` is useful for those running an instance locally for debugging purposes. The benefit as I see it is that it's all in one place.

I think we probably can remove the `Rakefile` totally if we wanted to and do everything in GitHub actions. It's a little harder to generate the site since you then have to know the specific commands to run to generate/build/serve but we can also document that somewhere.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @Rylan12!

> Each `Rakefile` task is pretty much just one line (either `brew generate-*-api` or `ruby script/generate-api-samples.rb`), so we could honestly scrap the `Rakefile` and do everything in GitHub actions.

Yeh, I think this is worth doing in a follow-up PR 👍🏻

PR #853, opened 2023-05-31
Fix analytics data

MikeMcQuaid:
- 30 day (InfluxDB) analytics data is shared on macOS and Linux
- Add `faraday-retry` gem so Jekyll stops complaining about it on startup
- Let `brew style --fix` do some reformatting


katrkatcloud (commented):
[](https://github.com/Homebrew/formulae.brew.sh/pull/853/files)

PR #858, opened 2023-06-06
config: remove feed

EricFromCanada:
As with Homebrew/brew#15501, this slightly speeds up generation and it's no longer needed as of Homebrew/brew.sh#957.


MikeMcQuaid:
Great fix @EricFromCanada, thanks!

PR #860, opened 2023-06-06
formula.html: fix grammar

p-linnane:
"because it unmaintained/deprecated upstream" ---> "because it is unmaintained/deprecated upstream"


Bo98:
Existing behaviour is consistent with brew: https://github.com/Homebrew/brew/blob/e9bfcb929625ef4c050cd1b79399fa926c31c745/Library/Homebrew/formula_installer.rb#L199-L215

The problem will be because we're missing the mapping: https://github.com/Homebrew/brew/blob/e4dab6384372daad531855cd56c5c3d8bf7501bb/Library/Homebrew/deprecate_disable.rb#L10-L23

PR #872, opened 2023-06-16
Gemfile: use newer Jekyll.

MikeMcQuaid:
Jekyll 4 is compatible with the GitHub Pages deployed from GitHub Actions pipeline we're using and generates about 3x faster on my machine.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,5 +3,9 @@
 source "https://rubygems.org"
 
 gem "faraday-retry"
-gem "github-pages", group: :jekyll_plugins
+gem "jekyll"
+gem "jekyll-redirect-from"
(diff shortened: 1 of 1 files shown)


Bo98:
Surprised you didn't need at least the SCSS change from: https://github.com/Homebrew/brew.sh/pull/934/files#diff-cd886530abfa495dc6782189ca779f1304bb73bc1758e5d3f0968448dfb67abe

I did when I tested out locally. Maybe some quirk with how `remote_theme` works.

MikeMcQuaid:
@Bo98 Weird! Will happily accept that change alone in Homebrew/brew.sh or can make it myself, whatever's easier for you.

PR #892, opened 2023-07-02
workflows/scheduled: remove cask update-reset

Bo98:
`brew update-reset` does not exit with a failure code if the fetch fails, so a stale JSON could be generated in those scenarios (see https://github.com/Homebrew/brew/issues/15616). Replace the logic with https://github.com/Homebrew/actions/pull/382.

Also add some timeout values in case it prevents https://github.com/Homebrew/formulae.brew.sh/actions/runs/5322147907. Maybe not but no harm trying.

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -19,6 +19,7 @@ jobs:
     if: startsWith( github.repository, 'Homebrew/' )
     name: Generate homebrew/cask data, pages and API
     runs-on: macos-latest
+    timeout-minutes: 20
     steps:
       - name: Check out repository
         uses: actions/checkout@main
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
> `brew update-reset` does not exit with a failure code if the fetch fails, so a stale JSON could be generated in those scenarios (see [Homebrew/brew#15616](https://github.com/Homebrew/brew/issues/15616)).

Should probably fix that?

MikeMcQuaid:
(Thanks for the PR though, makes sense!)

Bo98:
> Should probably fix that?

Probably. I think it makes some sense to continue to try reset if the fetch fails, but it should remember the failing exit code.

MikeMcQuaid:
> I can see the value of continuing to try reset if the fetch fails, but it should at least remember the failing exit code.

Agreed 👍🏻

PR #906, opened 2023-07-04
workflows/scheduled: skip core/cask updates where possible

Bo98:
Means less likely for failure if there's remote issues.

Requires https://github.com/Homebrew/actions/pull/391.


MikeMcQuaid (approved):
Makes sense to me, thanks again @Bo98!

PR #926, opened 2023-07-08
workflows/schduled: temporarily use previous analytics data

Bo98:
This may take some time to sort so unblock API generation for now by using the previous analytics data.


Bo98:
Should have opened this on a non-fork.

PR #949, opened 2023-08-10
Support URL sharding

ZhongRuoyu:
Instead of constructing the formula path manually, let's use `ruby_source_path` instead.

Fixes https://github.com/orgs/Homebrew/discussions/4704.


SMillerDev (approved):

MikeMcQuaid (commented):
Great work thanks @ZhongRuoyu!

PR #993, opened 2023-12-18
Add deprecated/disabled messages to cask pages

Rylan12:
Follow-up to https://github.com/Homebrew/brew/pull/16292

Show a warning when casks are deprecated or disabled now.

---

Something interesting that I noticed (that affects both formulae and casks), the `title` attribute of the `h2` includes the deprecation reason, but it just uses the pre-set reason instead of converting it to the actual string.

For example, take `zdoom` (which will soon contain `deprecate! date: "2023-12-17", because: :discontinued`). Right now, the `title` attribute is set to:

```
This cask has been deprecated since 2023-12-17 because it discontinued
```

Instead of:

```
This cask has been deprecated since 2023-12-17 because it is discontinued upstream
```

Anyway, something for another PR

### DIFF
--- a/_layouts/cask.html
+++ b/_layouts/cask.html
@@ -5,7 +5,16 @@ permalink: :title
 {%- assign token = page.title -%}
 {%- assign formula_path = "formula" -%}
 {%- assign c = site.data.cask[token] -%}
-<h2>{{ token }}</h2>
+<h2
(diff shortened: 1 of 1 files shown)


Bo98:
Yeah that issue affects formulae too.

Bo98 (approved):

MikeMcQuaid:
Great work thanks @Rylan12 and thanks @Bo98 for merging.

PR #1031, opened 2024-01-17
[WIP] Start using internal formula json

apainintheneck:
This is the slimmer JSON that is missing certain keys that won't be missed. The goal is that this smaller file contains the same amount of information as the previous file but means less information is flying around the internet and it will help speed up JSON parsing times as well since there's less to parse.

Related to:
- https://github.com/Homebrew/brew/pull/16460
- https://github.com/Homebrew/brew/issues/16410

### DIFF
--- a/.github/workflows/scheduled.yml
+++ b/.github/workflows/scheduled.yml
@@ -74,7 +74,7 @@ jobs:
           HOMEBREW_DEVELOPER: 1
 
       - name: Archive data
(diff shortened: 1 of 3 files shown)


apainintheneck:
I'm still trying to think of the best way to test this and beyond that https://github.com/Homebrew/brew/pull/16460 needs to land before this can get merged in.

MikeMcQuaid (approved):
Looks reasonable, thanks @apainintheneck! Agreed about not merging yet.

apainintheneck:
Closing as this approach is no longer needed after the changes to https://github.com/Homebrew/brew/pull/16460. The code will end up being simpler.

PR #1032, opened 2024-01-17
fix: CI use ruby 3.1

SMillerDev:

PR #1033, opened 2024-01-17
Use Ruby 3.1.4

MikeMcQuaid:
- use a `.ruby-version` file so this version is used both locally and in CI by ruby/setup-ruby@v1
- ensure we use the Bundler cache everywhere
- re-add a `Gemfile.lock` because it actually makes sense again with our new approach to deploying to GitHub Pages
- re-add `bundler` to the `dependabot.yml` because it's now needed again

Closes https://github.com/Homebrew/formulae.brew.sh/pull/1032


Gemfile.lock
```diff
@@ -0,0 +1,96 @@
...
+    safe_yaml (1.0.5)
+    sass-embedded (1.58.3-arm64-darwin)
+      google-protobuf (~> 3.21)
+    terminal-table (3.0.2)
+      unicode-display_width (>= 1.1.1, < 3)
+    unicode-display_width (2.4.2)
+    webrick (1.8.1)
+
+PLATFORMS
+  arm64-darwin-23
```
Bo98, line 84: Try this:

```
bundle lock --add-platform x86_64-darwin
bundle lock --remove-platform arm64-darwin-23
bundle lock --add-platform arm64-darwin
bundle lock --add-platform x86_64-linux
bundle lock --add-platform arm-linux
bundle lock --add-platform aarch64-linux
```

and if it works to make it more generic:

```
bundle lock --add-platform ruby
```

Can also just the `ruby` platform if you prefer, though will build from source always.

.github/dependabot.yml
```diff
@@ -1,11 +1,19 @@
...
-      - dependency-name: actions/stale
-      - dependency-name: dessant/lock-threads
+- package-ecosystem: bundler
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10
+  allow:
+  - dependency-type: direct
+  - dependency-type: indirect
```
Bo98, line 10:
```suggestion
  - dependency-type: all
```

I think this works - or at least we do this in `brew`.

Gemfile.lock
```diff
@@ -0,0 +1,113 @@
...
+    terminal-table (3.0.2)
+      unicode-display_width (>= 1.1.1, < 3)
+    unicode-display_width (2.4.2)
+    webrick (1.8.1)
+
+PLATFORMS
+  aarch64-linux
+  arm-linux
+  arm64-darwin
+  arm64-darwin-23
```
Bo98, line 99:
```suggestion
  arm64-darwin
```

Bo98 (approved):

PR #1055, opened 2024-01-25
docs/api: drop non-functioning `--all`

p-linnane:
`--all` is not a valid option for `brew info`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -11,7 +11,7 @@ redirect_from:
 
 ### List metadata for all {{ site.taps.core.name }} formulae or {{ site.taps.cask.name }} casks
 
-List the `brew info --json --all` output for all current {{ site.taps.core.fullname }} formulae or {{ site.taps.cask.fullname }} casks.
+List the `brew info --json` output for all current {{ site.taps.core.fullname }} formulae or {{ site.taps.cask.fullname }} casks.
 
 ```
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -11,7 +11,7 @@ redirect_from:
 
 ### List metadata for all {{ site.taps.core.name }} formulae or {{ site.taps.cask.name }} casks
 
-List the `brew info --json --all` output for all current {{ site.taps.core.fullname }} formulae or {{ site.taps.cask.fullname }} casks.
+List the `brew info --json --eval-all` output for all current {{ site.taps.core.fullname }} formulae or {{ site.taps.cask.fullname }} casks.
```
MikeMcQuaid, line 14:
```suggestion
List the `brew info --json` output for all current {{ site.taps.core.fullname }} formulae or {{ site.taps.cask.fullname }} casks.
```

p-linnane: This returns `Error: Invalid usage: this command requires a formula or cask argument`

MikeMcQuaid: Yeh, I don't think this needs to be an executable command and, pretty soon, it'll require a dedicated command anyway.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @p-linnane!

PR #1069, opened 2024-02-04
Start signing internal v3 json

apainintheneck:
Now that the PR on the main brew repo got merged in we're generating the internal v3 JSON but we're not storing or using it anywhere. This updates the scheduled generation task to not only store the resulting JSON but also sign it to allow us to use it as a reference when building the loader code.

- https://github.com/Homebrew/brew/pull/16541
- https://github.com/Homebrew/brew/issues/16410


MikeMcQuaid:
Thanks @apainintheneck!

PR #1081, opened 2024-02-22
script/generate-api-samples: add retries

Bo98:
Make this less likely to fail.

Longer term, this probably shouldn't download anything and should just generate something locally.


SMillerDev (approved):

MikeMcQuaid:
Makes sense, thanks @Bo98!



> Longer term, this probably shouldn't download anything and should just generate something locally.

Agreed 👍🏻

PR #1095, opened 2024-03-11
_includes/cask: update cask table layout

bevanjkay:
This PR updates the table layout on - https://formulae.brew.sh/cask/ To match the same layout as formulae - https://formulae.brew.sh/formula/

Related: https://github.com/Homebrew/brew/issues/16876

### DIFF
--- a/_includes/cask.html
+++ b/_includes/cask.html
@@ -6,5 +6,5 @@
             {{ include.token }}
 {%- endif -%}
         </td>
-        <td>{{ include.cask.name.first }}</td>
         <td>{{ include.cask.version | truncate: 25 }}</td>
+        <td>{{ include.cask.desc | xml_escape }}</td>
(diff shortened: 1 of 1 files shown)


SMillerDev (approved):

EricFromCanada:
Including the cask name along with the token was intentional, since a cask token does not always include the full name of the package that a user might be searching for. If including the description, it should be added as a fourth column, IMO.

_includes/cask.html
```diff
@@ -6,5 +6,5 @@
             {{ include.token }}
 {%- endif -%}
         </td>
-        <td>{{ include.cask.name.first }}</td>
         <td>{{ include.cask.version | truncate: 25 }}</td>
+        <td>{{ include.cask.desc | xml_escape }}</td>
```
MikeMcQuaid, line 10:
```suggestion
        <td>{{ include.cask.desc | xml_escape }}</td>
        <td>{{ include.cask.name.first }}</td>
```

MikeMcQuaid (commented):
Thanks @bevanjkay and @EricFromCanada!

PR #1105, opened 2024-03-20
Revert "Revert "Revert "Revert "workflows/scheduled: temporarily use previous analytics data""""

Bo98:
I've spent too many hours on this and I ultimately think it's unfixable on our side. It's clear something very suddenly happened (like a server code deploy) for things to suddenly cease to function properly on larger datasets and may require reaching out to support. https://github.com/Homebrew/homebrew-formula-analytics/pull/433

So let's get this necromance this revert train before the previous data expires after 24 hours, and to restore API updates.


MikeMcQuaid:
Thanks @Bo98, makes sense for now.

PR #1115, opened 2024-03-31
Use SVG logo in header

leoheitmannruiz:
As in https://github.com/Homebrew/brew.sh/pull/931


MikeMcQuaid:
Thanks @leoheitmannruiz!

PR #1145, opened 2024-04-27
update cask token handling

razvanazamfirei:
This changes the way we are handling the token. Currently, for casks including `@`, we see the following:
```html
<tr>
        <td><a href="/cask/brave-browserbeta">brave-browserbeta</a></td>
        <td>1.66.90.0</td>
        <td>Web browser focusing on privacy</td>
        <td>Brave Beta</td>
</tr>
``` 
instead of: 
```html
<tr>
        <td><a href="/cask/brave-browser@beta">brave-browser@beta</a></td>
        <td>1.66.90.0</td>
        <td>Web browser focusing on privacy</td>
        <td>Brave Beta</td>
</tr>
```

### DIFF
--- a/_includes/cask.html
+++ b/_includes/cask.html
@@ -1,9 +1,9 @@
 
         <td>
-{%- if site.data.cask[include.token] != nil -%}
-            <a href="{{ site.baseurl }}/cask/{{ include.token | uri_escape }}">{{ include.token | escape }}</a>
+{%- if site.data.cask[include.sort_key] != nil -%}
+            <a href="{{ site.baseurl }}/cask/{{ include.cask.token | uri_escape }}">{{ include.cask.token | escape }}</a>
 {%- else -%}
-            {{ include.token | escape }}
+            {{ include.cask.token | escape }}
 {%- endif -%}
         </td>
         <td>{{ include.cask.version | truncate: 25 | escape }}</td>
--- a/cask_index.html
+++ b/cask_index.html
@@ -11,9 +11,9 @@ permalink: /cask/
     {%- assign sorted_casks = site.data.cask | sort -%}
     {%- for c in sorted_casks -%}
     <tr>
-        {%- assign token = c[0] -%}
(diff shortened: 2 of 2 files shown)


_includes/cask.html
```diff
@@ -1,10 +1,7 @@
 
         <td>
-{%- if site.data.cask[include.token] != nil -%}
-            <a href="{{ site.baseurl }}/cask/{{ include.token | uri_escape }}">{{ include.token | escape }}</a>
-{%- else -%}
-            {{ include.token | escape }}
-{%- endif -%}
```
EricFromCanada, line 7: Retain this if statement; the non-URL version is used for lists of conflicting casks that may not be in `homebrew/cask`, e.g. https://formulae.brew.sh/cask/1password-cli.
Also, retain the use of `include.token` for the if statement, but use `include.cask.token` for output. This is because the key in the list of `sorted_casks` isn't actually the token (that should probably be renamed), it's a slightly modified version with special characters changed or removed.

cask_index.html
```diff
@@ -11,9 +11,8 @@ <h2><a href="{{ site.baseurl }}/api/cask.json"><code>/api/cask.json</code> (JSON
     {%- assign sorted_casks = site.data.cask | sort -%}
     {%- for c in sorted_casks -%}
     <tr>
-        {%- assign token = c[0] -%}
         {%- assign cask = c[1] -%}
-        {%- include cask.html token=token cask=cask -%}
```
EricFromCanada, line 16: As above: ideally, rename this to "key" or something to suggest it's not always the same as the token value.

EricFromCanada (changes requested):

razvanazamfirei:
Thanks, @EricFromCanada! I made the changes.

EricFromCanada (approved):

EricFromCanada:
Nice work @razvanazamfirei!

MikeMcQuaid:
Thanks @razvanazamfirei and @EricFromCanada for review/merge!

PR #1149, opened 2024-04-30
fix missing data in versioned cask pages

EricFromCanada:
As Liquid doesn't allow characters like `@`, `.`, or `+` in key names, we need to filter those out of cask token names as we did for formulae. More PRs are forthcoming, but this fixes the most obvious issue.


MikeMcQuaid:
Thanks @EricFromCanada!

PR #1151, opened 2024-05-01
fixes for versioned casks' API and analytics

EricFromCanada:
Follow-up to #1149 with more versioned cask fixes.


MikeMcQuaid:
Makes sense, thanks again @EricFromCanada!

PR #1152, opened 2024-05-01
drop usage of `formula_path`

EricFromCanada:
The `formula-linux` path has been unused since the linuxbrew merge, so no need to keep this logic around.


MikeMcQuaid:
Thanks again @EricFromCanada!

PR #1153, opened 2024-05-03
rework setup of cask includes

EricFromCanada:
This reworks how includes work for cask data (and fixes some missing conflicting cask links) by:

- rewording variables to distinguish between "token/tokens" (cask token strings), "data_token" (cask tokens adjusted to [match Jekyll's sanitization](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/readers/data_reader.rb#L70-L73)), "cask/casks" (full key+value info from site data), and "cdata" or "c" (cask value data whose properties can be fetched with dot notation),
	- variables in includes have "include_" prepended to avoid lurking bugs with (lack of) variable scope
- changing `includes/casks.html` to infer a data_token from each of its passed tokens and pass those to `includes/cask.html`,
- changing `includes/cask.html` to expect a data_token and a token, and do its own cask data retrieval if available. This is because it can be called by either:
  - `/cask_index.html`, which is looping through all the site's cask data, or
  - `includes/casks.html`, which can only provide a token which may or may not be for a cask whose data is in the site, e.g. for a cask page that lists cask dependencies or conflicts not in homebrew/cask.

This also replaces an if…for…else with a simpler [for…else](https://shopify.github.io/liquid/tags/iteration/#else) in `_layouts/cask_json.json`.

### DIFF
--- a/_includes/cask.html
+++ b/_includes/cask.html
@@ -1,11 +1,15 @@
 
         <td>
-{%- if site.data.cask[include.sort_key] != nil -%}
-            <a href="{{ site.baseurl }}/cask/{{ include.cask.token | uri_escape }}">{{ include.cask.token | escape }}</a>
+{%- assign include_cdata = site.data.cask[include.data_token] -%}
+{%- if include_cdata != nil -%}
+            <a href="{{ site.baseurl }}/cask/{{ include.token | uri_escape }}">{{ include.token | escape }}</a></td>
+        <td>{{ include_cdata.version | truncate: 25 | escape }}</td>
+        <td>{{ include_cdata.desc | escape }}</td>
+        <td>{{ include_cdata.name.first | escape }}
 {%- else -%}
-            {{ include.cask.token | escape }}
+            {{ include.token | escape }}</td>
+        <td></td>
+        <td></td>
+        <td>
 {%- endif -%}
         </td>
-        <td>{{ include.cask.version | truncate: 25 | escape }}</td>
-        <td>{{ include.cask.desc | escape }}</td>
-        <td>{{ include.cask.name.first | escape }}</td>
--- a/_includes/casks.html
+++ b/_includes/casks.html
@@ -1,10 +1,10 @@
-{%- if include.casks.size > 0 %}
+{%- if include.tokens.size > 0 %}
 <p>{{ include.description }}:</p>
 <table>
-    {%- for token in include.casks -%}
+    {%- for include_token in include.tokens -%}
     <tr>
-        {%- assign cask = site.data.cask[token] -%}
-        {%- include cask.html token=token cask=cask -%}
+        {%- assign include_data_token = include_token | remove: "@" | remove: "." | replace: "+", "_" -%}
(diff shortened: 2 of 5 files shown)


_includes/cask.html
```diff
@@ -1,11 +1,15 @@
 
         <td>
-{%- if site.data.cask[include.sort_key] != nil -%}
-            <a href="{{ site.baseurl }}/cask/{{ include.cask.token | uri_escape }}">{{ include.cask.token | escape }}</a>
+{%- assign inc_c = site.data.cask[include.data_token] -%}
+{%- if inc_c != nil -%}
+            <a href="{{ site.baseurl }}/cask/{{ include.token | uri_escape }}">{{ include.token | escape }}</a></td>
+        <td>{{ inc_c.version | truncate: 25 | escape }}</td>
+        <td>{{ inc_c.desc | escape }}</td>
+        <td>{{ inc_c.name.first | escape }}
```
MikeMcQuaid, line 8:
```suggestion
{%- assign include_cask = site.data.cask[include.data_token] -%}
{%- if include_cask != nil -%}
            <a href="{{ site.baseurl }}/cask/{{ include.token | uri_escape }}">{{ include.token | escape }}</a></td>
        <td>{{ include_cask.version | truncate: 25 | escape }}</td>
        <td>{{ include_cask.desc | escape }}</td>
        <td>{{ include_cask.name.first | escape }}
```

unsure of any of these `include.token` should be `include_cask`?

_includes/casks.html
```diff
@@ -1,10 +1,10 @@
-{%- if include.casks.size > 0 %}
+{%- if include.tokens.size > 0 %}
 <p>{{ include.description }}:</p>
 <table>
-    {%- for token in include.casks -%}
+    {%- for inc_token in include.tokens -%}
```
MikeMcQuaid, line 4:
```suggestion
    {%- for include_cask_token in include.tokens -%}
```

_includes/casks.html
```diff
@@ -1,10 +1,10 @@
...
     <tr>
-        {%- assign cask = site.data.cask[token] -%}
-        {%- include cask.html token=token cask=cask -%}
+        {%- assign inc_data_token = inc_token | remove: "@" | remove: "." | replace: "+", "_" -%}
+        {%- include cask.html data_token=inc_data_token token=inc_token -%}
```
MikeMcQuaid, line 7:
```suggestion
        {%- assign include_cask_data_token = include_cask_token | remove: "@" | remove: "." | replace: "+", "_" -%}
        {%- include cask.html data_token=include_cask_data_token token= include_cask_token -%}
```

MikeMcQuaid (approved):
Looks good! I think some naming could be a little nicer here but not blocking, up to you.

_includes/cask.html
MikeMcQuaid: ok, thanks. do whatever you think is best :+1:

MikeMcQuaid (approved):
Looks good thanks @EricFromCanada!

PR #1154, opened 2024-05-03
rework setup of formula includes

EricFromCanada:
These changes mirror those from #1153 for how formulae are handled, and also:
- changes `_layouts/cask.html` so that references to non-core casks in conflicts_with aren't skipped
- adds to `_includes/cask.html` the handling of references to old names of casks (depends on Homebrew/brew#17225)
- drops the number of characters to display of versions to 20
- removes a FUSE requirement reference

### DIFF
--- a/formula_index.html
+++ b/formula_index.html
@@ -10,10 +10,11 @@ redirect_from: /formula-linux/
 
 <table>
     {%- assign sorted_formulae = site.data.formula | sort -%}
-    {%- for f in sorted_formulae -%}
+    {%- for formula in sorted_formulae -%}
     <tr>
-        {%- assign formula = f[1] -%}
-        {%- include formula.html formula=formula -%}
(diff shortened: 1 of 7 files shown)


formula_index.html
```diff
@@ -10,10 +10,11 @@ <h2><a href="{{ site.baseurl }}/api/formula.json"><code>/api/formula.json</code>
 
 <table>
     {%- assign sorted_formulae = site.data.formula | sort -%}
-    {%- for f in sorted_formulae -%}
+    {%- for formula in sorted_formulae -%}
     <tr>
-        {%- assign formula = f[1] -%}
-        {%- include formula.html formula=formula -%}
+        {%- assign data_fname = formula[0] -%}
+        {%- assign fname = formula[1].name -%}
```
EricFromCanada, line 16: 🏎

PR #1160, opened 2024-05-07
import & display full deprecate/disable reasons

EricFromCanada:
This adds deprecated/disabled badges lists of casks, and copies over the deprecate/disable reasons from [`deprecate_disable.rb`](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/deprecate_disable.rb) and inserts either the full text [corresponding with the reason symbol](https://docs.brew.sh/Deprecating-Disabling-and-Removing-Formulae#deprecate-and-disable-reasons) into the hovertext or the text from the formula/cask itself if customized.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -58,6 +58,26 @@ analytics:
       - name: Build Errors
         path: build-error
 
+# sync with Library/Homebrew/deprecate_disable.rb
+reasons:
+  formula:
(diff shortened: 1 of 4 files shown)


SMillerDev (approved):

MikeMcQuaid (approved):

PR #1173, opened 2024-05-13
list cask fonts on separate page

EricFromCanada:
Lists any cask fonts on https://formulae.brew.sh/cask-font/. Each font's page is still under /cask/; this just keeps the listing page lengths manageable.


MikeMcQuaid (approved):
Makes sense, thanks again @EricFromCanada!

PR #1190, opened 2024-05-23
update ignored paths

EricFromCanada:
- `.gitignore`: add `api/internal/` and sort contents
- `Rakefile`: add paths in `CLOBBER.include` lines, plus minor fixes


MikeMcQuaid (approved):
Thanks again @EricFromCanada!

PR #1285, opened 2024-07-13
Synchronize shared configuration

p-linnane:
The workflow created the branch, but never opened a PR for this.


Bo98:
yeah the workflow is spamming the PR API too much so it never succeeds fully as it hits GitHub's spam limits. Does work on a re-run however.

Bo98 (approved):

p-linnane:
Makes sense, thanks!

PR #1317, opened 2024-07-31
Update workflow filename in README

muzimuzhi:
Functionality of `workflows/scheduled.yml` was moved into `workflows/tests.yml` in 7bce4cc0d5d7 Make workflows/scheduled replace workflows/tests. (2024-05-24).


p-linnane (approved):
Thanks @muzimuzhi!

PR #1414, opened 2024-09-23
workflows/tests: use `create-or-update-issue` action

ZhongRuoyu:
This was added in Homebrew/actions#588. It simplifies the workflow, and reduces the number of API calls.


MikeMcQuaid (approved):
So much nicer, thanks @ZhongRuoyu!

PR #1765, opened 2025-03-13
Idea: show recommended replacement

botantony:
It would be nice to show what people can use instead of disabled/deprecated package.

Unfortunately I cannot test changes locally (`rake generate` throws `Error: No InfluxDB credentials found in HOMEBREW_INFLUXDB_TOKEN!`, `bundle install` needs different Ruby version: `Your Ruby version is 3.4.2, but your Gemfile specified 3.3.7`)

It would be nice to add styles to `<hr/>` tag

### DIFF
--- a/_includes/replacement.html
+++ b/_includes/replacement.html
@@ -1,0 +1,1 @@
+<p class="recommended-replacement">Recommended replacement: <strong><a href="{{ site.baseurl }}/{{ include.r_type }}/{{ include.item | uri_escape }}">{{ include.item | escape }}</a></strong></p>
--- a/_layouts/cask.html
+++ b/_layouts/cask.html
@@ -16,6 +16,12 @@ permalink: :title
     {%- endif -%}
 </h2>
 
+{%- if c.disable_replacement %}
+{%- include replacement.html item=c.disable_replacement r_type="cask" %}
+{%- elsif c.deprecation_replacement %}
+{%- include replacement.html item=c.deprecation_replacement r_type="cask" %}
+{%- endif %}
+
 {%- include install_command.html item=c.token modifier=" --cask" %}
 <p class="names">Name{%- if c.name.size > 1 -%}s{%- endif -%}:
 {%- for name in c.name %}
--- a/_layouts/formula.html
+++ b/_layouts/formula.html
@@ -16,6 +16,12 @@ permalink: :title
     {%- endif -%}
 </h2>
 
+{%- if f.disable_replacement %}
+{%- include replacement.html item=f.disable_replacement r_type="formula" %}
+{%- elsif f.deprecation_replacement %}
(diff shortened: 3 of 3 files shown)


_layouts/cask.html
```diff
@@ -16,6 +16,16 @@
...
 
+{%- if c.disable_replacement %}
+<p class="recommended-replacement">Recommended replacement: <strong><a href="{{ site.baseurl }}/cask/{{ c.disable_replacement | uri_escape }}">{{ c.disable_replacement | escape }}</a></strong></p>
+{%- include install_command.html item=c.disable_replacement modifier=" --cask" %}
+<hr/>
+{%- elsif c.deprecation_replacement %}
+<p class="recommended-replacement">Recommended replacement: <strong><a href="{{ site.baseurl }}/cask/{{ c.deprecation_replacement | uri_escape }}">{{ c.deprecation_replacement | escape }}</a></strong></p>
+{%- include install_command.html item=c.deprecation_replacement modifier=" --caks" %}
+<hr/>
+{%- endif %}
```
SMillerDev, line 27: Do we know for sure that the replacement is a cask? It could just as well be a formula

botantony: I tried to fix this but it looks... meh...
Most likely I'd make it into a separate `_includes` module (the same as `install_command.html`)

botantony:
Here's how it looks:
image

With some simple styling it may looks like this:
image

_includes/replacement.html
```diff
@@ -0,0 +1,14 @@
...
+{%- else %}
+{%- assign r_type = "cask" %}
+{%- endif %}
+<div>
+<p class="recommended-replacement">Recommended replacement: <strong><a href="{{ site.baseurl }}/{{ r_type }}/{{ include.item | uri_escape }}">{{ include.item | escape }}</a></strong></p>
+{%- if r_type == "cask" %}
+{%- include install_command.html item=include.item modifier=" --cask" %}
+{%- else %}
+{%- include install_command.html item=include.item %}
+{%- endif %}
```
EricFromCanada, line 12: Omit the suggested install command. Leave it to the user to follow the link to the replacement and copy the install command from there.

_includes/replacement.html
```diff
@@ -0,0 +1,6 @@
+{%- if site.data.formula[include.item] %}
+{%- assign r_type = "formula" %}
+{%- else %}
+{%- assign r_type = "cask" %}
+{%- endif %}
```
EricFromCanada, line 5: You can skip this logic by passing "formula" or "cask" as another parameter, e.g.:
```
{%- include replacement.html item=f.disable_replacement type="formula" %}
```

MikeMcQuaid (approved):
Looks good, thanks again @botantony!

PR #1789, opened 2025-03-20
formula page: split Linux into its own section

EricFromCanada:
Now that the first ARM64 Linux bottles are available (e.g. https://formulae.brew.sh/formula/gnu-tar), split the bottle list into macOS on Apple Silicon, macOS on Intel, and Linux.


MikeMcQuaid (approved):
Looks great!

PR #1793, opened 2025-03-20
formula.html: do not separate bottle groups without bottles

botantony:
Right now `Bottle` section has separating `<tr>` tags regardless of existing bottles. Examples:

Only Linux bottles:
image

Only ARM macOS bottles:
image

Only macOS (Intel + ARM) bottles:
image

This PR fixes it (same pages in the respective order):
image

image

image


MikeMcQuaid (approved):
Thanks @botantony!

PR #1844, opened 2025-04-22
recommended replacement: adjust to new API

botantony:
In new version of Homebrew there's no `disable_replacement` and `deprecation_replacement` but separate fields for casks and formulae


MikeMcQuaid (approved):
Thanks!

Issue #1792, opened 2025-03-20
API samples generation job should not have a circular dependency on the site


Bo98:
## What you were trying to do (and why)

Recently we had an outage that could have been solved quicker if we were able to rebuild the site. However site builds have a dependency on the site itself being up and functional.

## What happened (include screenshots)

The API samples generation job fails because the website was down: https://github.com/Homebrew/formulae.brew.sh/actions/runs/13978393367/job/39137776970

## What you expected to happen

The API samples generation job does not have any dependency on the live site.

PR #1906, opened 2025-06-03
Remove circular dependency in API sample generation

Rylan12:
Fixes https://github.com/Homebrew/formulae.brew.sh/issues/1792

Let's just source these samples directly from the generated files. That way, they'll always work even if the site is down.

Note: for `formula.json`, it's a little more tricky because the generated file is actually a template that is only expanded after the site is generated. Instead, just modify the `formula/wget.json` file to match the syntax that would be used in the larger file (indent, add surrounding brackets, and remove the analytics).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -138,43 +138,8 @@ jobs:
           path: data-analytics.tar.gz
           retention-days: 1
 
-  generate-samples:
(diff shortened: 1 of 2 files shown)


MikeMcQuaid (approved):
Makes sense to me, thanks @Rylan12!

PR #1913, opened 2025-06-06
Don't generate API samples for Dependabot updates

Rylan12:
See https://github.com/Homebrew/formulae.brew.sh/pull/1911

We don't generate the analytics API data for dependabot PRs, so we can't generate the API samples either. This PR gates the API sample generation by the same condition used to generate the analytics API data and sign the API files.


EricFromCanada (approved):

MikeMcQuaid:
Thanks @Rylan12!

PR #1936, opened 2025-06-26
Add `executables.txt` from `Homebrew/command-not-found` to internal API

Rylan12:
See https://github.com/Homebrew/homebrew-command-not-found/issues/246

This PR pulls the latest `executables.txt` file to be served at `https://formulae.brew.sh/api/internal/executables.txt` for future use by `brew which-formula`.

Although as I'm opening this I'm wondering: is serving this from formulae.brew.sh better than having users download directly from GitHub?

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -177,6 +177,12 @@ jobs:
           tar xvf data-cask/data-cask.tar.gz
           tar xvf data-core/data-core.tar.gz
 
+      - name: Install oras for pulling from GitHub Packages
+        run: brew install oras
+
+      - name: Pull executables.txt from GitHub Packages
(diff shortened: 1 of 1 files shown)


Rylan12:
> Although as I'm opening this I'm wondering: is serving this from formulae.brew.sh better than having users download directly from GitHub?

If we have users download from GitHub directly, I think we need to perform two network calls: one to the manifest file to retrieve the sha256 of the blob, and then a second to actually download (like we do for formulae). So that might be a good reason to host it here.

MikeMcQuaid (approved):
Thanks! Think this is fine for now but let's avoid documenting this API anywhere until we switch over to a nicer, JSON-based approach.

PR #2001, opened 2025-09-02
Only cancel in-progress scheduled generation in pull requests

Rylan12:
See https://github.com/Homebrew/formulae.brew.sh/issues/2000#issuecomment-3246601507

We don't want to cancel in-progress generation jobs for concurrency reasons. Instead, let the previous job finish and simply add the new job to a queue. This will reduce the frequency of generation failures if the generation job happens to take a long time.

Although it's worth noting, this will make it harder to catch issues that result in API generation taking an unusually long time.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ on:
 
 concurrency:
   group: scheduled-${{ github.ref }}-${{ github.event_name }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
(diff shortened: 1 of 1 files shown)


SMillerDev (approved):

Bo98 (approved):

Bo98:
> Although it's worth noting, this will make it harder to catch issues that result in API generation taking an unusually long time.

We have timeouts everywhere so not really. If the current timeouts are too long we can shorten them.

p-linnane:
Thanks @Rylan12!

PR #2077, opened 2025-12-05
Add rel attribute to formula and cask pages

SMillerDev:
Inspired by https://x.com/elie2222/status/1996564439957647690


SMillerDev:
I disagree with the Copilot comments that the code of the cask is not an alternate for the page.

MikeMcQuaid:
Good idea, thanks!

PR #2136, opened 2026-02-23
workflows/tests: clean runner before formula API

cho-m:
Not sure yet if this actually works as expected. Mainly opening to discuss best way to handle:
- https://formulae.brew.sh/formula/openssl@1.1#default

The issue seems to be due to GitHub runners having a copy of removed formula in core tap:
- https://github.com/actions/runner-images/blob/main/images/macos/scripts/build/install-openssl.sh

I added `git clean -ffdx` similar to what **actions/checkout** does - https://github.com/actions/checkout?tab=readme-ov-file#usage
```
    # Whether to execute `git clean -ffdx && git reset --hard HEAD` before fetching
    # Default: true
    clean: ''
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,6 +71,8 @@ jobs:
           core: true
           cask: false
 
+      - run: brew test-bot --only-cleanup-before
+
       - run: brew generate-formula-api
         env:
           HOMEBREW_DEVELOPER: 1


cho-m:
Sadly didn't work. May need to uninstall it

https://github.com/Homebrew/formulae.brew.sh/actions/runs/22319058388/job/64572027748#step:6:23479
```
a formula/openssl@1.1.html
```

cho-m:
I think that worked. At least can no longer see any `openssl@1.1` in:
- https://github.com/Homebrew/formulae.brew.sh/actions/runs/22320085839/job/64575693944?pr=2136#step:6:1

MikeMcQuaid (approved):

PR #2173, opened 2026-04-12
Generate shared internal packages API

Rylan12:
Run `brew generate-internal-api` to include the `/api/internal/packages.TAG.json` endpoints

---

Also, while we're here, add `stable: true` so the API is regenerated using the latest `main` and not `stable`

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,7 @@ jobs:
         with:
           core: false
           cask: true
+          stable: false
 
       - run: brew generate-cask-api
         env:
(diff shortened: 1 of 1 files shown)


Rylan12:
Hmm interesting... after https://github.com/Homebrew/actions/pull/837, the API is always populated using the `stable` commit of `brew` and so it never has the latest `main` changes. I wonder if this is the right approach or not

MikeMcQuaid (commented):
Thanks!

MikeMcQuaid (approved):

PR #2196, opened 2026-05-12
workflows/tests: use unstable Homebrew for analytics.

MikeMcQuaid:
This is consistent with other, working workflows.


SMillerDev (approved):

SMillerDev:
> Your Ruby version is 4.0.2, but your Gemfile specified 4.0.3

And we're back here

MikeMcQuaid:
@SMillerDev Yeh, I think this is an issue with the action not updating portable Ruby properly before invoking it. A new PR for that 🔜

SMillerDev (approved):

PR #2254, opened 2026-08-03
Compact bottle/analytics tables on small screens

andrew:
The shared brew.sh stylesheet sets `display: block` on all table cells below 700px, which turns the bottle matrix and analytics tables on formula/cask pages into long narrow columns with mostly empty space beside them.

This adds a small local stylesheet (`assets/css/formulae.css`) with a `no-stack` class that keeps opted-in tables as real tables on narrow viewports, and applies it to the bottle, versions and analytics tables. It also reworks the analytics table so intervals are columns rather than a repeated header/row pair per category/interval, which reads better on desktop too.

**Bottle table (mobile, `a52dec`)**

| before | after |
|---|---|
| mobile-bottle-before | mobile-bottle-after |

**Analytics table (mobile, `a52dec`)**

| before | after |
|---|---|
| mobile-analytics-before | mobile-analytics-after |

**Analytics table (desktop, `ffmpeg`)**

| before | after |
|---|---|
| desktop-analytics-before | desktop-analytics-after |


_layouts/formula.html
```diff
@@ -291,26 +291,35 @@
...
-        <td><code>{{ fa.formula | escape }}</code></td>
-        <td class="number-data">{{ fa.count }}</td>
+        <th>{{ category.name }}
+            {%- assign variant_suffix = variant.formula | remove_first: fname | strip -%}
+            {%- if variant_suffix != "" %} (<code>{{ variant_suffix | escape }}</code>){% endif -%}
+        </th>
+        {%- for interval in site.analytics.intervals -%}
+            {%- assign interval_data = site.data.analytics[category.path].homebrew-core[interval.path] -%}
+            {%- assign match = interval_data.formulae[fname] | where: "formula", variant.formula | first -%}
+        <td class="number-data">{% if interval_data %}{{ match.count | default: "0" }}{% endif %}</td>
```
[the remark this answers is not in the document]
andrew: The empty cell here is intentional. The only case where `interval_data` is nil is build-error 90d/365d, where no data file is generated at all, so an empty cell is more accurate than `0`. Jekyll's `where` returns its input unchanged when it doesn't respond to `select`, so `nil | where: ... | first` safely evaluates to nil and the `{% if interval_data %}` guard on the output handles the rest.

_layouts/cask.html
```diff
@@ -165,21 +165,32 @@
...
-        <td><code>{{ fa.cask | escape }}</code></td>
-        <td class="number-data">{{ fa.count }}</td>
+        <th>Installs
+            {%- assign variant_suffix = variant.cask | remove_first: token | strip -%}
+            {%- if variant_suffix != "" %} (<code>{{ variant_suffix | escape }}</code>){% endif -%}
+        </th>
+        {%- for interval in site.analytics.intervals -%}
+            {%- assign match = site.data.analytics.cask-install.homebrew-cask[interval.path].formulae[token] | where: "cask", variant.cask | first -%}
+        <td class="number-data">{{ match.count | default: "0" }}</td>
+        {%- endfor -%}
```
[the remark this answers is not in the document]
andrew: `nil | where: ...` is safe in Jekyll (it returns nil), and `match.count | default: "0"` already emits `0` for the nil case. All three cask-install intervals are generated so this path isn't hit in practice anyway.

_layouts/base.html
```diff
@@ -57,6 +57,7 @@
     <link rel="icon" href="https://brew.sh/assets/img/homebrew.svg" sizes="any" type="image/svg+xml">
     <link rel="apple-touch-icon" href="https://brew.sh/assets/img/apple-touch-icon.png">
     <link rel="stylesheet" href="https://brew.sh/assets/css/style.css" type="text/css" media="screen">
+    <link rel="stylesheet" href="{{ site.baseurl }}/assets/css/formulae.css" type="text/css" media="screen">
```
SMillerDev: Can't this just be in the `style.css` above?

andrew: It can be, but that's in a different repo, so will need to split this in two

SMillerDev: I think that would be nicer to maintain.

andrew: Done — opened Homebrew/brew.sh#1346 and dropped the local stylesheet here in bc67ed02. This PR now depends on that one landing first.

andrew:
https://github.com/Homebrew/brew.sh/pull/1346 merged, this is ready for review again now.

SMillerDev (approved):
Thanks @andrew, it's a big improvement

SMillerDev:
Okay, no clue why the build keeps timing out after 20 minutes

andrew:
@SMillerDev I think it's my fault, investigating

PR #2270, opened 2026-08-18
Skip API generation when inputs are unchanged

p-linnane:
Caches each generate job's output so the 15-minute site builds skip regeneration when nothing changed: the advisory index is keyed on the advisories/ tree plus the generator source, analytics on the day. Advisory generation failures fail the build rather than republishing stale data; analytics keeps its restore-on-failure fallback, now only when a cache was actually restored, and both trees are sentinel-validated before archiving. Follows up on the suggestion in Homebrew/brew#23530.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -156,11 +156,41 @@ jobs:
           cask: true
           stable: false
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Love it!

PR #2276, opened 2026-08-23
Stop advertising Rosetta-only macOS versions as Intel

p-linnane:
Cask pages listed macOS 27 Golden Gate as an Intel platform, reported in https://github.com/orgs/Homebrew/discussions/7026.

Cask API data carries both `arm64_golden_gate` and an unprefixed `golden_gate`, the latter describing x86_64 installs through Rosetta. `cask.html` treated every unprefixed tag as an Intel platform, so pages advertised a macOS version that no Intel Mac runs, and general-purpose Rosetta support ends after macOS 27 anyway.

Filtering `golden_gate` by name would need editing again on every macOS release and would regress silently in between: the site rebuilds from freshly generated cask data every 15 minutes, so a new unprefixed tag reaches production without a commit. `_config.yml` instead lists the macOS versions that do run on Intel, and unprefixed tags outside it are dropped once, before the supported platforms table is counted and rendered, so future releases are hidden by default. Casks such as 1password now list Golden Gate only under Apple Silicon, and a cask whose only platform is Rosetta-only renders no table rather than an empty one. Apple Silicon, Linux and Intel through Tahoe are unchanged, and no generated API data is touched.

`script/validate-build.rb` only inspected JSON, so nothing checked the generated HTML. It now fails the build when the allowlist is missing or malformed, when no supported platforms table is inspected, when an architecture rowgroup stops being rendered, when rowgroup markup appears without the table selector matching, when a table is empty or a row cannot be inspected, when a `rowspan` disagrees with its row count, or when a platform outside the allowlist renders under Intel. Several of those exist so the check cannot pass vacuously after a later markup or configuration change.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -40,6 +40,17 @@ taps:
     fullname: Homebrew/homebrew-cask
     remote: https://github.com/Homebrew/homebrew-cask
 
+# macOS versions that run on Intel hardware. Unprefixed cask platform tags
+# outside this list are Rosetta-only and should not be advertised.
+intel_macos_platforms:
+  - tahoe
+  - sequoia
+  - sonoma
+  - ventura
+  - monterey
+  - big_sur
+  - catalina
+
 analytics:
   intervals:
     - name: 30 days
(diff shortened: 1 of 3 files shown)


SMillerDev (approved):

MikeMcQuaid (approved):
Thanks!

PR #2282, opened 2026-08-31
layout: use `bottles` CSS id for `Bottle` and `Supported platform` sections

botantony:
Add small border between `Supported platform` blocks, similar to how it works with formulae `Bottle` section.

Before:
image

After:
image

Closes https://github.com/orgs/Homebrew/discussions/7042

### DIFF
--- a/_layouts/cask.html
+++ b/_layouts/cask.html
@@ -163,7 +163,7 @@ permalink: :title
 {%- endfor %}
 
 <p>Supported platforms:</p>
-<table class="full-width no-stack">
+<table class="full-width no-stack" id="bottles">
     {%- if arm64_platform_count > 0 %}
     <tbody>
     {%- assign subsequent = false -%}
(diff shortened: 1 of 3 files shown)


SMillerDev:
I don't really see the difference. It think, if we want a border, it would be clearer to use css to create an actual border.

MikeMcQuaid (approved):
Thanks! Take a look at Copilot comments.

botantony:
> I don't really see the difference. It think, if we want a border, it would be clearer to use css to create an actual border.

I didn't want to touch CSS as it exists in a separate repo but it seems like I have to. I added
```scss
#bottles {
  border-collapse: collapse;

  tbody:not(:first-child) {
    border-collapse: separate;
    border-top: 10px solid rgba(0, 0, 0, 0.6);
  }
}
```
to the styles, `bottles` id to the table and this is what it looks like:

image

What do you think?

SMillerDev:
I think that looks much better, nice work

botantony:
https://github.com/Homebrew/brew.sh/pull/1359

